Issue #1, opened 2014-10-14
Link between backend and Revealjs / js/ css


cmoulliard:
This is an excellent idea to move to a separate project the asciidoctor backend of revealjs. Nevertheless, there is still an issue as there is no link between the version supported of revealjs and this backend.

I know that the purpose of this repo is not at all to maintain what revealjs did but that should be easier for our users that there is a git submodule mapped to a tag version of this backend or something else (bower, ruby-bower, ...).

This is what I have done with hyla project where revealjs is embedded (css, js) - https://github.com/cmoulliard/hyla/tree/master/lib/resources/revealjs and where there is a command that the user can use to create a project for a slideshow, .... including also a config.yaml file with parameters (https://github.com/cmoulliard/hyla/blob/master/documentation/scripts/generate_slideshow_revealjs_config.sh)

obilodeau:
I just left a comment in #10 that is strongly related to this issue. I suggest you go and read it and let me know what you think.

mojavelinux:
Since #10 is now closed, perhaps it's best to continue the conversation in this issue since it deals directly with the problem at hand.

obilodeau:
I see two use-cases defended here: 
* slides that are located inside an open source project and that should be maintained over time
* slides used in a presentation and that are frozen and in an archival state

I mostly deal with the latter type of slide decks. So I'll talk about that use case. We've patched reveal.js so that the speaker notes plugin output nicer notes and we've added a theme. The theme is branded so it won't be sent upstream and the notes patch was done in a hurry and can be seen as a hack. Additionally we used an unreleased asciidoctor back-end (before the -reveal.js repo) and we rendered our slides before some of our patches were merged.

For our presentation sources repo, we used git submodules to our own forks (both asciidoctor-reveal.js and revealjs) and created a Makefile that would build our slides. For [our public slides repo](https://github.com/eset/slides) we can pin specific version of the specific reveal.js back-end we used using git submodules again. This has the benefit that github pages does the right thing (tm) IOW it pulls the exact version of reveal.js we require and we can host slides there with a very lightweight repo.

I would like to hear how that could be achieved using Bower. Git submodules are not easy to use if you are not familiar with git.

cmoulliard:
You could also use Hyla project (
https://github.com/cmoulliard/hyla/blob/master/documentation/introduction.adoc#6-generate-a-slideshow)
as it supports natively Revealjs without the need to use bower, ...

[...]

PR #10, opened 2015-01-15
Warning about version 3 incompatibility

### LINKED ISSUE #9: Compatibility issue with reveal.js 3.0.0
With reveal.js the produced HTML is a static html page. I lost a couple of hours to discover that. At least it should be stated in the readme file until compatibility is set.

antoinesd:
Added a warning to prevent use of reveal.js 3.0.0

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -37,6 +37,8 @@ WARNING: We recommend using Slim 2.1.0 until https://github.com/asciidoctor/asci
 
   $ git clone git://github.com/hakimel/reveal.js.git
 
+WARNING: Currently asciidoctor-reveal.js is not compatible with reveal.js 3.0.0, be careful to clone or copy version 2.6.2 or older
+
 TIP: If you are using https://pages.github.com/[GitHub Pages], plan ahead by keeping your source files on `master` branch and all output files on the `gh-pages` branch.
 
 == Source Examples


README.adoc
```diff
@@ -37,6 +37,8 @@ WARNING: We recommend using Slim 2.1.0 until https://github.com/asciidoctor/asci
 
   $ git clone git://github.com/hakimel/reveal.js.git
 
+WARNING: Currently asciidoctor-reveal.js is not compatible with reveal.js 3.0.0, be careful to clone or copy version 2.6.2 or older
```
mojavelinux: Perhaps we should recommend cloning a specific branch in the `git clone` statement. wdyt?

    $ git clone -b 2.6.2 https://github.com/hakimel/reveal.js

Notice I also updated the clone URL to be https as I've heard this is much better for anyone behind a proxy.

We should also start thinking about recommending the use of Bower or another tool to get reveal.js rather than just git clone, but that's a topic for a separate issue.

kito99:
@mojavelinux , can you accept this request? I just lost about 30-40 minutes because I used reavel.js 3.0 ;-). 

By the way, you can manually edit the output to get it to work, or run grunt with the "css -- cssmin" and "js --uglify" targets to generate the minified files referenced by the HTML output. However, the output still references default.css as the theme,  regardless of what theme you select. In other words, I got it down to one step (replace default.css with the real theme CSS file).

obilodeau:
This raises the issue of deployment and reproducibility of being able to render the slide deck. I produced slides with a version that was still inside [asciidoctor/asciidoctor-backends](https://github.com/asciidoctor/asciidoctor-backends/tree/master/slim/revealjs) and with reveal.js 2.x (whatever last release before 3.0).

Since reveal.js' master was still moving and because we patched both asciidoctor and reveal.js (theming) we opted for git submodules inside our repository. The advantage of submodules being that the repo you refer to is pinned to a specific commit (the backend and reveal.js). Thus reproducible slides.

I think the reproducibility issue would be affected if we did something like what is suggested in #1 (and I realize now that I maybe should have commented there). How can one get reproducible slide decks if bower fetches always the latest version of everything?

I can document a git submodule approach but if someone thinks it can be done better with other tools suggest something. The git submodule approach requires more git knowledge for sure.

mojavelinux:
I like the approach of using Bower because then we are working off releases, which is going to be more stable.

To support multiple versions of reveal.js, we can create separate branches in this repository and treat them a bit like separate projects. Though I'm fine with just dropping reveal 2.x, which would allow us to deal with the branches later.

mojavelinux:
I'm going to merge this while we sort out how to move forward so others don't get stuck in the interim.

mojavelinux:
Please feel free to continue the discussion.

mojavelinux:
> How can one get reproducible slide decks if bower fetches always the latest version of everything?

Bower should be configured to fetch specific versions. That's very possible.

PR #11, opened 2015-02-02
Adding ability to change reveal.js path

JGrenier:
When generating more than one document it should be nice to be able pointing the same folder.
Actually, the reveal.js dir is hardcoded in the document.html.slim and the folder has to be copied in each presentation folder..

addition of a revealjsdir attribute to specify the location of the reveal.js folder.
Based on the work of @dsyer for deckJs asciidoctor/asciidoctor-backends@1b28d98

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -181,6 +181,10 @@ Default settings are based on `reveal.js` default settings.
 |<file\|URL>
 | Overrides CSS with given file or URL. Default is disabled.
 
+|:revealjsdir:
+|<file\|URL>
+| Overrides reveal.js directory. Example : ../reveal.js
+
 |:revealjs_controls:
 |*true*, false
 |Display controls in the bottom right corner.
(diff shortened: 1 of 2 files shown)


mojavelinux:
Excellent! Looks good to me. @obilodeau and @robertpanzer, any comments or good to merge?

robertpanzer:
Great!
Would it be possible now to refer to the reveal.js resources on Github as well?

mojavelinux:
Yep, I suppose so.

robertpanzer:
That's awesome if you want to publish slides online!

mojavelinux:
Merged! Thanks @JGrenier!

obilodeau:
The only thing I'm concerned with this PR is that it doesn't use the _ separator like all the other config flag uses. Introducing consistency problems within the backend.

Would anyone be offended if I fixed that? IOW: s/revealjsdir/revealjs_dir/

mojavelinux:
@obilodeau, true, but `revealjsdir` is consistent with how we do directory settings in other parts of Asciidoctor, so I'd rather keep it this way.

Keep in mind that in the future, we are planning to make the hyphen and/or underscore in attribute names insignificant, so you'd be able to type `revealjs-dir` or `revealjs_dir` and still get the same result.

obilodeau:
Fair enough

PR #12, opened 2015-02-21
Upgrade to reveal 3.0.0

danhyun:
Use https://github.com/hakimel/reveal.js/releases/tag/3.0.0 as a guide

As of 3.0.0 they excluding minified files from the distributed release artifact. If we want minification we'll have to do it ourselves.

Other than a few minor changes this was straightforward.

### DIFF
--- a/templates/slim/document.html.slim
+++ b/templates/slim/document.html.slim
@@ -10,12 +10,12 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     meta content="yes" name="apple-mobile-web-app-capable"
     meta content="black-translucent" name="apple-mobile-web-app-status-bar-style"
     meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, minimal-ui" name="viewport"
-    link href="#{revealjsdir}/css/reveal.min.css" rel="stylesheet"
+    link href="#{revealjsdir}/css/reveal.css" rel="stylesheet"
     / Default theme required even when using custom theme
     - if attr? :revealjs_customtheme
       link rel='stylesheet' href=(attr :revealjs_customtheme) id='theme'
     - else
-      link rel='stylesheet' href='#{revealjsdir}/css/theme/default.css' id='theme'
+      link rel='stylesheet' href='#{revealjsdir}/css/theme/black.css' id='theme'
     - if attr? :stem
       - asset_uri_scheme = (attr 'asset-uri-scheme', 'https')
       - asset_uri_scheme = %(#{asset_uri_scheme}:) unless asset_uri_scheme.empty?
@@ -36,7 +36,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
           },
           TeX: {#{eqnums_opt}}
           });
-      script src='#{cdn_base}/mathjax/2.4.0/MathJax.js?config=TeX-MML-AM_HTMLorMML'  
+      script src='#{cdn_base}/mathjax/2.4.0/MathJax.js?config=TeX-MML-AM_HTMLorMML'
     - case attr 'source-highlighter'
     - when 'coderay'
       - if (attr 'coderay-css', 'class') == 'class'
@@ -67,7 +67,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
             p: small=author
         =content
     script src="#{revealjsdir}/lib/js/head.min.js"
-    script src="#{revealjsdir}/js/reveal.min.js"
+    script src="#{revealjsdir}/js/reveal.js"
     javascript:
       // See https://github.com/hakimel/reveal.js#configuration for a full list of configuration options
       Reveal.initialize({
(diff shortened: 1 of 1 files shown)


pvdissel:
:+1:

obilodeau:
Looked at the patch, didn't test.

If this is all that there is to it then great! You get my +1. Bonus points if you can link to slides produced by the toolchain. ;)

obilodeau:
Forgot to mention that we should consider discussing online publishing workflow so that stuff produced with previous asciidoctor-reveal.js / previous reveal.js can still be built. Let's start this discussion in #10.

pvdissel:
I use it in my Spock Framework presentation, its repository can be found here:
- https://github.com/pvdissel/talks
But I have not yet stored the generated output publicly, as it's not finished yet.

I would love to be able to unit/integration test this. I've looked at https://github.com/asciidoctor/asciidoctor-doctest, but didn't spend to much time into trying to set it up.. therefore skipped it. But would be good to have doctests setup

mojavelinux:
> we should consider discussing online publishing workflow so that stuff produced with previous asciidoctor-reveal.js / previous reveal.js can still be built.

For now, let's create a reveal.js-2.x branch prior to merging this fix. Once that's done, I'm all for moving forward. I'll be using reveal.js 3.x for my presentation this week, so I'll give it a thorough test.

mojavelinux:
> But would be good to have doctests setup

The person to talk to for that is @jirutka. He recently helped get doctest setup for the LaTeX backend, so you can also look there for an example (there are recent discussions on the topic).

templates/slim/document.html.slim
```diff
@@ -129,7 +129,8 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
             { src: '#{revealjsdir}/plugin/markdown/markdown.js', condition: function() { return !!document.querySelector( '[data-markdown]' ); } },
             #{(attr? 'source-highlighter', 'highlight.js') ? "{ src: '#{revealjsdir}/plugin/highlight/highlight.js', async: true, callback: function() { hljs.initHighlightingOnLoad(); } }," : nil}
             { src: '#{revealjsdir}/plugin/zoom-js/zoom.js', async: true, condition: function() { return !!document.body.classList; } },
-            { src: '#{revealjsdir}/plugin/notes/notes.js', async: true, condition: function() { return !!document.body.classList; } }
+            { src: '#{revealjsdir}/plugin/notes/notes.js', async: true, condition: function() { return !!document.body.classList; } },
+            { src: '#{revealjsdir}/plugin/math/math.js', async: true}
```
mojavelinux: I don't think we want to load the math.js extension since we are handling it ourselves (or we need to take out our own handling...but I think some configuration is needed, so better to do it ourselves I think).

templates/slim/document.html.slim
```diff
@@ -10,12 +10,12 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
...
     meta content="black-translucent" name="apple-mobile-web-app-status-bar-style"
     meta content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, minimal-ui" name="viewport"
-    link href="#{revealjsdir}/css/reveal.min.css" rel="stylesheet"
+    link href="#{revealjsdir}/css/reveal.css" rel="stylesheet"
     / Default theme required even when using custom theme
     - if attr? :revealjs_customtheme
       link rel='stylesheet' href=(attr :revealjs_customtheme) id='theme'
     - else
-      link rel='stylesheet' href='#{revealjsdir}/css/theme/default.css' id='theme'
+      link rel='stylesheet' href='#{revealjsdir}/css/theme/black.css' id='theme'
```
mojavelinux: Any reason we are switching to the black theme by default?

pvdissel: Black is the new default, it's true :wink: 
But serious, the default theme isn't there anymore in 3.0 (https://github.com/hakimel/reveal.js/tree/master/css/theme), black is now the default

mojavelinux: Oh, I see. reveal.js no longer has default. Then +1

mojavelinux:
Thanks for bringing us to reveal.js 3.0, @danhyun!

I made some minor updates to the settings and docs, then merged manually as c0becc21d4ff9d70210922577e26e84babe98e4c.

mojavelinux:
Here's the reveal.js-2.x branch to use with older presentations.

https://github.com/asciidoctor/asciidoctor-reveal.js/tree/reveal.js-2.x

The modifications to the AsciiDoc are very minimal. Just some of the options for the theme and transition changed. Other than that, using reveal.js just works as far as I can tell.

templates/slim/document.html.slim
mojavelinux: Thanks @pvdissel. I updated the README.adoc accordingly.

Btw, it seems that "league" is the new name for what used to be the default theme...in case you want to use it. I've decided on "league" as the default since it will look familiar to those trying it the first time.

jirutka:
I’m quite busy now, but I’ll try to look at reveal.js 3.0 soon and help with DocTest setup. However, it’s HTML backend, so it should be very easy.

jirutka:
DocTest is in place, look at branch `doctest`.

Now it’s necessary to fix two blocking errors [here](https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/slim/block_preamble.html.slim#L6) and [here](https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/slim/block_toc.html.slim#L12).

It seems that these templates are _very_ outdated. How this backend actually differs from plain HTML5 backend (except `document.html.slim`)? It’d be perhaps easier to take current [HTML5/Slim](https://github.com/asciidoctor/asciidoctor-backends/tree/master/slim/html5) templates and modify it for reveal.js, instead of updating these templates.

Anyway, after fixing and updating the templates, someone should generate testing examples (`bundle exec rake generate`), review them and then commit to the repo.

danhyun:
@mojavelinux we should also probably update `README`

> Currently revealjs uses a rather old version of highlight.js that does not handle callouts correctly. To fix this download a current version of highlight.js and copy it to reveal.js/plugin/highlight/highlight.js.

reveal.js 3.0.0 uses hightlight.js 8.2

https://github.com/hakimel/reveal.js/blob/master/plugin/highlight/highlight.js

So we should probably remove the `NOTE:` and update slim templates to set highlight.js as default source highlighter.

Let me know what you think, I can do the work for this.

mojavelinux:
Yes, let's proceed!

PR #13, opened 2015-03-01
Make used highlight style configurable

pvdissel:
By default, the 'zenburn' syntax highlighting style from highlight.js
is included with reveal.js and used. But now you can provide
your own path|URL to a custom syntax highlighing style.
eg. one of the many other styles highlight.js offers.

Also corrected the link to the reveal.js theme edit/creation documentation

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -181,6 +181,10 @@ Default settings are based on `reveal.js` default settings.
 |<file\|URL>
 | Overrides CSS with given file or URL. Default is disabled.
 
+|:highlightjs_style:
+|<file\|URL>
+| Overrides https://highlightjs.org[highlight.js] CSS style with given file or URL. Default is disabled.
+
 |:revealjsdir:
 |<file\|URL>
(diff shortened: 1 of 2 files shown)


obilodeau:
Sound reasonable but didn't test. +1

pvdissel:
I use it in my Spock Framework presentation, its repository can be found here:
- https://github.com/pvdissel/talks
But I have not yet stored the generated output publicly, as it's not finished yet.

I would love to be able to unit/integration test this. I've looked at https://github.com/asciidoctor/asciidoctor-doctest, but didn't spend to much time into trying to set it up.. therefore skipped it. But would be good to have doctests setup

mojavelinux:
@danhyun agreed. I'll fix it post-merge.

mojavelinux:
Thanks @pvdissel!

mojavelinux:
Please note that the attribute is now `highlightjs-theme` to be consistent with the HTML backend. (Though it behaves slightly differently because it treats it as a file path).

PR #15, opened 2015-03-04
Update slim template to look for `highlightjs` to be consistent with Asciidoctor documentation

### LINKED ISSUE #14: highlight.js missing from HTML output
I'm not sure what I'm doing wrong, but the output for my file does not include the highlight.js dependency. As far as I can tell, this should be there by default. I'm using reveal.js 2.62. These are the only dependencies I'm seeing:

 ```javascript
  dependencies: [
      { src: 'reveal.js/lib/js/classList.js', condition: function() { return !document.body.classList; } },
      { src: 'reveal.js/plugin/markdown/marked.js', condition: function() { return !!document.querySelector( '[data-markdown]' ); } },
      { src: 'reveal.js/plugin/markdown/markdown.js', condition: function() { return !!document.querySelector( '[data-markdown]' ); } },
      
      { src: 'reveal.js/plugin/zoom-js/zoom.js', async: true, condition: function() { return !!document.body.classList; } },
      { src: 'reveal.js/plugin/notes/notes.js', async: true, condition: function() { return !!document.body.classList; } }
```

danhyun:
Fixes #14 

Update slim template to look for `highlightjs` to be consistent with Asciidoctor documentation
http://asciidoctor.org/docs/user-manual/#source-code-blocks

### DIFF
--- a/templates/slim/document.html.slim
+++ b/templates/slim/document.html.slim
@@ -36,7 +36,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
           },
           TeX: {#{eqnums_opt}}
           });
-      script src='#{cdn_base}/mathjax/2.4.0/MathJax.js?config=TeX-MML-AM_HTMLorMML'  
+      script src='#{cdn_base}/mathjax/2.4.0/MathJax.js?config=TeX-MML-AM_HTMLorMML'
     - case attr 'source-highlighter'
     - when 'coderay'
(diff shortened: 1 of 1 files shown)


mojavelinux:
Looks good! Thanks!

Note that the value `highlight.js` is still supported in Asciidoctor core, but we've deprecated it to avoid the confusion of having two values.

PR #28, opened 2015-05-05
Added width / scale / margins options

### LINKED ISSUE #26: Add options for presentation size
Reveal.js has 4 options related to [presentation size](https://github.com/hakimel/reveal.js#presentation-size), which I have found very useful for large screen / projector sizes where you can easily fit a big image and a few lines of text beneath it, by setting `width` and `height` to larger values. It would be useful to have at least those two options, and possibly also the `minScale` and `maxScale` options as well.

wackywendell:
Fixes #26. The changes are pretty straightforward, and follow the convention above of defaulting to the reveal.js defaults.

I didn't add any testing because I don't really know how to do automated testing of something like this, but I did make some slides where I used these features and it worked for me.


mojavelinux:
Looks great! Thanks!

>  I don't really know how to do automated testing of something like this

We'll eventually use asciidoctor-doctest, but it's not setup yet.

Issue #18, opened 2015-03-18
Drop the data- prefix on the slide background attributes


mojavelinux:
Reveal.js recognizes a handful of data attributes (attributes with the prefix `data-`) for setting the background properties of a slide. Currently, the Asciidoctor section template passes these attributes through verbatim. I don't think we need the `data-` prefix and therefore propose to drop it.

The `data-` prefix is required in the HTML as a namespace. We don't need this prefix in the AsciiDoc document as it doesn't add any semantic value nor is is required for compliance.

Before

```
[data-background-image=evil-minion.png]
== Lab results
```

After

```
[background-image=evil-minion.png]
== Lab results
```

For backwards compatibility, we could allow both, but we should definitely only document the attribute names without the `data-` prefix.

obilodeau:
+1

obilodeau:
This will be fixed by PR #52 which will get merged once reviewed. Closing.

PR #38, opened 2015-07-09
Drop the data- prefix on the slide attributes

andreabedini:
Copy and pasting from #18 

> Reveal.js recognizes a handful of data attributes (attributes with the prefix data-) for setting the background properties of a slide. Currently, the Asciidoctor section template passes these attributes through verbatim. I don't think we need the data- prefix and therefore propose to drop it.
>
>  The data- prefix is required in the HTML as a namespace. We don't need this prefix in the AsciiDoc document as it doesn't add any semantic value nor is is required for compliance.

I also removed the prefix for the `transition` attribute.

I am not too familiar with asciidoctor templates. Is this all it takes?

### DIFF
--- a/templates/slim/section.html.slim
+++ b/templates/slim/section.html.slim
@@ -2,7 +2,7 @@
 - hide_title = (title = self.title) == '!'
 - if @level == 1 && !(subsections = sections).empty?
   section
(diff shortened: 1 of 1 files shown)


andreabedini:
I'll close this in favour of #52 which seems more thorough.

PR #39, opened 2015-07-09
Add reveal.js jade templates

rahmanusta:


mojavelinux:
Looks good to me! @obilodeau, these templates are a clone of the slim templates for use in tooling based on Asciidoctor.js like AsciiDocFX. We'll also be able to use them in the Chrome and Firefox extension (in theory).

rahmanusta:
Thank you @mojavelinux ! We can see the practice here :) 

https://github.com/asciidocfx/asciidoctor.js-reveal-demo

obilodeau:
Cool work. @mojavelinux do you know if there's a tilt jade back-end that would work with asciidoctor (I'm lost in all of this ruby templating) or if we'll have to maintain two template engine for the time being?

@rahmanusta a small update to the README to document the new template set, the fact that it works with asciidoctor.js and asciidocfx and links to these projects would be awesome. After this, its a merge.

Thanks

README.adoc
```diff
@@ -160,6 +160,10 @@ Here is https://github.com/hakimel/reveal.js#markup[upstream documentation] on t
 
 NOTE: The `reveal.js` backend only supports level 1 (`==`) and level 2 (`===`) sections. Level 3 sections are not supported and will not render correctly.
 
+== About Jade Templates
+
+`/templates/jade` directory contains jade template files they are ported from `/templates/slim` templates. This templates written to support reveal.js backend for Asciidoctor.js environment that is currently using in https://github.com/asciidocfx/AsciidocFX[AsciidocFX] editor. You can look at the https://github.com/asciidocfx/asciidoctor.js-reveal-demo[demo].
```
wackywendell: > This templates written

"These templates are written"?

rahmanusta: Of course, sorry for my bad english. I guess you can fix it :)

wackywendell: No problem! And don't worry about it - I know how difficult it is to learn a foreign language, and I'm always impressed by non-native speakers who can make real contributions to stuff written in English!

rahmanusta: Thank you @wackywendell for your understanding

README.adoc
```diff
@@ -160,6 +160,10 @@ Here is https://github.com/hakimel/reveal.js#markup[upstream documentation] on t
 
 NOTE: The `reveal.js` backend only supports level 1 (`==`) and level 2 (`===`) sections. Level 3 sections are not supported and will not render correctly.
 
+== About Jade Templates
+
+`/templates/jade` directory contains jade template files they are ported from `/templates/slim` templates. These templates written to support reveal.js backend for Asciidoctor.js environment that is currently using in https://github.com/asciidocfx/AsciidocFX[AsciidocFX] editor. You can look at the https://github.com/asciidocfx/asciidoctor.js-reveal-demo[demo].
```
wackywendell: > These templates written

Sorry to push again, but I think you are still missing a word: "These templates **are** written" (or "**were** written").

rahmanusta: Yes, I amended the commit.

mojavelinux:
Good question, @obilodeau. I made some progress using the tilt-jade gem from the Ruby side (server-side) along the following preload script:

```ruby
require 'tilt'
require 'tilt-jade/template'

class TiltJade::Template
  alias :original_render :render

  def render scope = Object.new, locals = {}, &block
    res = original_render scope, locals, &block
    ExecJS.compile(%(#{JadeJs::Source.contents}\n#{res})).eval 'anonymous({}, {})'
  end
end

Tilt.register TiltJade::Template, 'jade'
```

That at least gets the following to work:

```ruby
puts Tilt.new('test.jade').render
```

However, at this point we run into a mismatch between Ruby and JavaScript since the Jade engine relies on a variable scope inside the JavaScript engine. I'm not sure how to push Ruby variables down to JavaScript short of using Opal to build some sort of bridge.

At this point, I don't think it's possible to use the Jade templates from Ruby (outside Asciidoctor.js) with some additional hacking. There doesn't seem to exist any library that treats Jade purely as a server-side template language...which is the exact opposite problem we have using Slim inside of JavaScript. In other words, we're the pioneers unfortunately.

Another approach here is to setup a test suite that uses a JavaScript engine to test the templates. Perhaps we can integrate something of the sort into doctest.

mojavelinux:
I don't think it's a huge challenge to maintain two sets of templates. It's not really that many files and since the goal is to simply keep them in sync, we can just be mindful of that when making changes to the repository.

If there were some way to use the same set of templates in both Ruby and JavaScript, I would be recommending that we align, but sadly that problem just hasn't been solved yet.

mojavelinux:
Seems like mustache may be our best bet if the goal is to have a single template language for both Ruby and JavaScript. However, that also means we'd have to give up Slim, which I'm not keen on doing just yet.

mojavelinux:
Another option is to use Slm instead of Jade. That at least makes the templates more similar, thus easier to maintain. @rahmanusta, would you consider trying Slm instead of Jade to see if it works? If it does, I think it might be best to use Slm instead of Jade to ease maintenance. I'm open to your input.

mojavelinux:
https://github.com/slm-lang/slm

rahmanusta:
I see this conversation too late :) Yes, we can give it a try for slm. But, I think for now, we took much more effort with jade :) What is the downside of jade for this work?

mojavelinux:
As I mentioned, the benefit of slm is that the templates are closer to the original Slim templates and thus easier to maintain (using side-by-side comparison).

Another approach is to get the template converter working with Asciidoctor.js so that we can use the `asciidoctorjs` command to test the templates from the console. I think what @obilodeau and I are looking for is a way to validate changes to the templates without any external tool...and eventually create a doctest for them. Otherwise, we're just merging blind.

Mogztter:
We should also provide an embedded template. See https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/slim/embedded.html.slim

rahmanusta:
@Mogztter How is embedded templates usage?

I will write it but to test, can you give me an usage example for that?

mojavelinux:
@rahmanusta Basically it's when something else provides the HTML "shell" (basically the html and head tags) and the converter just makes the body part. It saves you from having to "cut out" the body content after conversion to get the part you need. Most of the time it just passes through the children, but in some cases it may need to create some extra things. In this case, probably the title slide.

rahmanusta:
I almost understand it. It reminds me of "head_footer" attribute while converting HTML. How can I test/activate embedded templates? 

Also, I wrote embedded template for reveal.js but, I didnt understand https://github.com/asciidoctor/asciidoctor-deck.js/blob/master/templates/haml/embedded.html.haml#L9 line for deck.js.

mojavelinux:
>  It reminds me of "head_footer" attribute while converting HTML.

Yes, that's what controls whether the top-level template is "document" or "embedded". In other words, there are two "root" templates from which to select that is controlled by that option.

That's just a fancy way of appending text to the next line as a way to deal with whitespace. Probably better to look at the original template to understand it: https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L274-L281

Regardless, I don't think we've sorted out how footnotes fit into slides yet, so I'd actually just leave that part off for now.

rahmanusta:
@mojavelinux I pushed embedded.jade for both templates.

Mogztter:
\o/

mojavelinux:
Nice. Thanks!

rahmanusta:
@Mogztter What happend to the process ? :)

mojavelinux:
I'm going to be getting back into this soon. I've been on a detour working on Bespoke.js, which will have very similar needs. But since it aligns with a contract I'm working on, it will give me an opportunity to look at and review the infrastructure for JavaScript-based templates in general.

rahmanusta:
+1 Dan. I also preworked for bespoke and resides it's own branch, it is very nice slide framework.

Mogztter:
@rahmanusta Sorry I've been busy lately and didn't find the time to work on this
@mojavelinux Good news!

obilodeau:
@rahmanusta I would prefer to have a similar template language since I'm already having a hard time learning to work with ruby and slim. Adding javascript/jade would just make contributing less fun.

That said I don't want to throw away all of your work so as a compromise if you maintain (and test it) it I'm open to merge it. The jade templates are behind by a few commits I've made in the last week. Also the data-background rework PR #52 should be some more work for you coming soon. 

Interested in porting the significant changes we do in slim to jade?

rahmanusta:
@obilodeau I can maintain and keep up update jade templates.

obilodeau:
@rahmanusta sorry for the hiatus on this. Can you fix the merge conflict in the README and catch up with the few changes that happened since your work? I'll merge quickly afterwards. Promise.

Thanks for your work.

mojavelinux:
I had a chance to use Jade to create several presentations (before I started work on the Bespoke.js converter). Jade feels similar enough to Slim that the concepts mostly map 1-to-1. While a single template language across platforms would be nice, it's just not something that exists yet.

I think the right first step is to merge the Jade templates and make sure that someone is looking after them. @rahmanusta, I'm happy that you have volunteered to do this, at least in the near term. Ideally, over time, other people will come forward. We should always be sure someone is around looking after them. Tests (doctests) would help massively here.

I think merging will open the door to a conversation about whether Jade templates is the best approach for Asciidoctor.js support. To insulate from future changes, I do believe that the converter should hide the fact that it's using the templates (for most users). I've suggested an approach in the Bespoke.js converter (see https://github.com/asciidoctor/asciidoctor-bespoke/blob/master/lib/asciidoctor-bespoke/converter.rb). Not only does introducing a concrete converter allow templates to be loaded from the Ruby install path, it also abstracts away which templates are loaded.

Which brings me to my final point. If we want ultimate flexibility, we'd just switch to writing the converter in Ruby instead of using templates. The upside is that we get translation to JavaScript (via Opal) for free. The downside is that it's harder for users to customize. With the templates, users can just clone and hack on them. If done in Ruby, then they either have to stay in Ruby or figure out how the make the equivalent template to start hacking on (and extra step). This is where I'm stuck in my thinking.

--

(@rahmanusta the concrete converter would be a good place to put the code you are using to load the templates from Asciidoctor.js, at least until it is available in Asciidoctor.js).

mojavelinux:
@obilodeau @rahmanusta Could either of you rebase so this starts as a single commit?

rahmanusta:
Thank you @obilodeau @mojavelinux . I rebased the works as a single commit. I'll be happy to see it merged. I'll look at the asciidoctor-bespoke, it may be good to integrate it to AsciidocFX.

mojavelinux:
Thanks @rahmanusta!

I'd love to see Bespoke.js support in AsciidocFX. I'm still flushing out some of the templates, so some things may change, but I'm nearing an alpha release.

obilodeau:
@rahmanusta if I use AsciidocFX, it seems to be using an older version of our back-end. Is there a way I can change a setting or use an environment variable to say use `~/src/asciidoctor-reveal.js.git/templates/jade` instead of the default one? I guess I can copy files over AsciidocFX's `conf/slide/templates/revealjs/` but it would be nice to be able to override for development purposes.

Unrelated, I saw that [jade was renamed pug](https://github.com/jadejs/jade). Do we want to rename everything before the merge or wait a bit for the new name to spread?

obilodeau:
Ok, this has waited long enough, I'm moving forward and I'll open a new issue to track the remaining things.

rahmanusta:
@obilodeau It is not configurable for now but we should make it configurable

PR #40, opened 2015-07-21
Update to reveal.js 3.1.0 themes

rwinch:
- Update list of available themes to match current release
- Update default to be black (default theme of current release)

Also note that in reveal.js-2.6.x there is no css file named league. The
default theme is named default. This means currently no theme is actually
loaded.

This commit fixes the theme for 3.1.x. Unfortuantely, the black theme does
not exist until 3.0.0. This means this commit will break 2.x version of
reveal.js.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -171,7 +171,7 @@ NOTE: Default settings are based on `reveal.js` default settings.
 |Attribute |Value(s) |Description
 
 |:revealjs_theme:
-|beige, black, blood, *league*, moon, night, serif, simple, sky, solarized, white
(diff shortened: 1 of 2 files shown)


obilodeau:
We have a separate branch for reveal.js 2.x support so it's fine to merge
this as is. Assuming the code is alright ;-) since I couldn't review it
because I'm offline.

mojavelinux:
This looks correct to me.

It's strange, I thought I remembered fixing this problem, but it could have been from some hotel room trying to get ready for a presentation at some point...and I forgot to follow through with a PR :) Thanks @rwinch!

PR #48, opened 2015-09-18
fix unnecessary nested section in vertical sections

lifei:
`====` is rendered to nested section which is not correct.

### DIFF
--- a/templates/slim/section.html.slim
+++ b/templates/slim/section.html.slim
@@ -11,7 +11,11 @@
       =subsection.convert
 / standalone slides
 - else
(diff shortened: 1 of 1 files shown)


mojavelinux:
Thanks for this fix.

I think we could use a dynamic tag element to make this more concise. Something like:

```
*{tag: %(h#{@level > 1 ? 2 : 1})=title
```

Could you see if you can get the title output down to one line in each case?

lifei:
You are right. I'm new to Ruby and slim. I do some modifications following your advice.

obilodeau:
@lifei will you update your branch based on @mojavelinux's recommendations?

If not, it's ok I'll do it. Just let me know.

lifei:
yes, I'll update the branch soon. thanks for remind.

lifei:
done. please review the branch.

mojavelinux:
At first glance that looks very reasonable to me. Could you remove the blank lines added in document.html.slim as they are not part of this change? Thanks!

lifei:
yeah, sure

lifei:
donw

mojavelinux:
Thanks!

Issue #43, opened 2015-07-30
Image in title slide is displayed in every slide


Zlika:
Hi,
If I put an image in the title slide, it gets printed in every slides! (at the top of the slide).

```
include::_settings_reveal.adoc[]

= My title
My name

image:https://upload.wikimedia.org/wikipedia/commons/thumb/3/35/Tux.svg/2000px-Tux.svg.png[width=150]
```

obilodeau:
This is due to bug #49. Until we fix it, as an ugly workaround, you can always use an empty title slide and skip it before anyone sees it.

```
= _

== My title

[.small]
My name

image:https://upload.wikimedia.org/wikipedia/commons/thumb/3/35/Tux.svg/2000px-Tux.svg.png[width=150]
```

obilodeau:
With current master the following worked fine for me. Reopen the issue if necessary.

```asciidoc
= My title
My name

image:https://upload.wikimedia.org/wikipedia/commons/thumb/3/35/Tux.svg/2000px-Tux.svg.png[width=150]

== My slide 1

My content
```

mojavelinux:
+1

PR #50, opened 2015-10-05
document preamble shown only on title slide (fixes #49, #43)

### LINKED ISSUE #49: document preamble (text after title before first section) is shown on every slide
A bug present in both deck.js and reveal.js related to the fact that block_preamble is processed after the title slide and will slip the content through inside a `<div>` which will not be properly nested inside the title's `<section>` element.

Already discussed here: http://discuss.asciidoctor.org/Using-Preambel-with-deck-js-scrambles-output-td2154.html

I don't think the solution of having the preamble in its own slide to be the proper one with regards to what one would expect when writing slides. I tried and failed to implement the second solution mentioned in the discussion.

@mojavelinux I tried listing attributes and can't find any attribute to match that would say: "this is the first slide". A section number would do. Also, it is unclear to me how to grab content of another block from within another template. Help appreciated :)

obilodeau:
As discussed in #49. Thanks for the guidance @mojavelinux. 

`@document.find_by` doesn't work on asciidoctor previous to 1.5.0 so I thought I mentioned minimum requirements in the README.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -287,3 +287,7 @@ Defaults to none
 |===
 
 If you want to build a custom theme or customize an existing one you should look at the https://github.com/hakimel/reveal.js/blob/master/css/theme/README.md[reveal.js documentation] and use the `revealjs_customtheme` AsciiDoc attribute to activate it.
+
+== Minimum Requirements
+
+* asciidoctor 1.5.0
--- a/templates/slim/block_preamble.html.slim
+++ b/templates/slim/block_preamble.html.slim
@@ -1,0 +1,2 @@
+/ file intentionally left blank
+/ preamble is shown on the title slide which is rendered by document template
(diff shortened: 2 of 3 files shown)


mojavelinux:
Agreed. Now that GitHub is on 1.5.0, I think that's the minimum version of Asciidoctor the ecosystem should support.

mojavelinux:
Maybe add a comment line in preamble.html.slim that explains the file is intentionally left blank?

mojavelinux:
Btw, nice work!

obilodeau:
Yes, I have no problem with the required version bump.

I added the comment. Good idea.

mojavelinux:
Looks excellent!

obilodeau:
I introduced a problem when there is no preamble.

```
undefined method `length' for nil:NilClass
  Use --trace for backtrace
```

obilodeau:
I expected the result of find_by to be an empty array and not a nil. Fixed by a straight push to master (sorry about that): 3daa174

mojavelinux:
The return value is fixed in the upcoming 1.5.3, but we'll have to live
with the nil guard in modules for now.

PR #51, opened 2015-10-06
Helpful back-end development tips

obilodeau:
Hopefully not offending anyone with duplicating some content here for greater offline availability, also known as plane coding ;)


mojavelinux:
Not at all. Moving content closer to where it is needed is never a bad thing :)

PR #34, opened 2015-05-16
resolves #7 support background options on title slide

mojavelinux:
* support background options on title slide
* use h1 for slide title instead of h2

PR #52, opened 2015-10-06
data-background attributes rework

obilodeau:
Integrated the work in #38 and the ideas I proposed in #34 for background image support as block image attributes. The PoC is there but the documentation is missing, I'll be waiting on feedback. There are suboptimal and non-rubyish things in there since I'm not familiar with that language.

Breaking changes:

* `data-background-image` is now `background` referring to the CSS attribute that reveal.js supports
* similarily, `data-transition`, `data-transition-speed`, `data-background-size`,  `data-background-repeat`, `data-background-transition` all lost their data- prefix since it was not required and seen as redundant (see #38 and #18)
* `:imagesdir:` is now honored on background attribute making image handling more consistent

TODO:
* [x] in addition to `background`, `canvas` should be supported when looking for background images
* [x] doc: background image URLs work
* [x] doc: background is now relative to :imagesdir: if set
* [x] doc: attributes supports whatever CSS supports
* [x] document new syntax format
* [x] link to https://github.com/hakimel/reveal.js/#slide-backgrounds, https://github.com/hakimel/reveal.js/#slide-transitions
* [x] ~~link to the relevant moz doc attributes since reveal.js hasn't documented them~~ upstream did it

I pushed my test slides so you can try to reproduce in your environments: 
* with section attributes: https://github.com/obilodeau/my-asciidoctor-tests/blob/master/data-background-oldstyle.adoc
* with new block image background attribute syntax: https://github.com/obilodeau/my-asciidoctor-tests/blob/master/data-background-newstyle.adoc

If accepted, I'll merge Dan's title-slide work and apply the same block_image syntax too.

Edit: added elements to doc TODO


templates/slim/block_image.html.slim
```diff
@@ -1,10 +1,11 @@
-.imageblock(id=@id class=[@style,role]
-    style=[("text-align: #{attr :align}" if attr? :align),("float: #{attr :float}" if attr? :float)].compact.join('; '))
-  .content
-    - if attr? :link
-      a.image href=(attr :link)
+- unless attr(1) == 'background'
```
obilodeau: this patch looks larger than it should due to diff being confused. all I did is inject this unless statement.

templates/slim/section.html.slim
```diff
@@ -1,8 +1,37 @@
-/ support for vertical slides
+/ OPTIONS PROCESSING
+/ extracting block image attributes to find an image to use as a background attribute
+- data_background, data_background_size, data_background_repeat,
+  data_background_transition = nil
+/ FIXME would have expected the following to work... which would probably be faster
+/- slide_images = find_by(:context => :image) {|image| image.attr(1) == 'background' }
+- slide_images = find_by(:context => :image)
```
obilodeau: Instead of iterating on all of the section's images shouldn't I be able to prefilter the results? I tried but I failed. cc @mojavelinux

obilodeau:
I forgot backwards compatibility. Any pointers about how to do it without hurting readability too much?

obilodeau:
@mojavelinux do you think the premises for that change are sound? If so, I would finish up the docs and merge this.

As a reminder, have a look at the [old markup](
https://github.com/obilodeau/my-asciidoctor-tests/raw/master/data-background-oldstyle.adoc) (that will still work) and the [new proposed markup](https://github.com/obilodeau/my-asciidoctor-tests/raw/master/data-background-newstyle.adoc) to get a feel of what this PR proposes.

An additional benefit I just noticed is that you see the background images for slides in github's asciidoc preview (since they are image macros instead of hidden away in section attributes).

nicolaiparlog:
I did not look deeply into the changes but I like the PR's direction (especially the `:imagesdir:` inclusion). Any ETA on when this might get merged?

obilodeau:
I'm waiting on the approval of contributors and users.

templates/slim/section.html.slim
mojavelinux: Yes, that should work. However, the was a bug in earlier versions of Asciidoctor so I believe it requires Asciidoctor 1.5.4.

templates/slim/block_image.html.slim
mojavelinux: You can write this as `(attr :alt)` instead of `(attr 1)`. The first positional attribute is always mapped to the `alt` attribute for images.

mojavelinux: Giving it some more thought, if you stick with (attr 1), then it still would be possible to provide alt text using the `alt` attribute explicitly. So now I'm inclined to say keep it how it is.

mojavelinux:
Nice work!

> An additional benefit I just noticed is that you see the background images for slides in github's asciidoc preview (since they are image macros instead of hidden away in section attributes).

That's exactly the idea. Visible content should be content. Attributes should modify.

Overall, I'd say it looks good. I tested it and I can confirm it works.

At first, I was a little reserved about overloading the alt text to promote the image to the background. In the Bespoke converter, I'm using a role for this purpose:

```
image::cover.jpg[Background image,1024,640,role=cover canvas]
```

or as two lines:

```
[.cover.canvas]
image::cover.jpg[Background image,1024,640]
```

Giving it some more thought, I realize that the alt text is usually superfluous and just getting in the way. Taking the approach you are proposing, my example could be simplified to:

```
image::cover.jpg[canvas,1024,640,size=cover]
```

That's probably a lot cleaner. I could still provide alt text if necessary:

```
image::cover.jpg[canvas,1024,640,size=cover,alt=Background image]
```

I tend to prefer the word "canvas" over "background" since it indicates the layer to which the image is applied as a background. (The canvas being the whole viewport).

So, after letting this sink in, I like the approach. My only change request is to use "canvas" instead of "background", though if you think background fits better for reveal.js users, then I'll let you make that call.

obilodeau:
Since `background` is the upstream revealjs term I would keep it but I think we can support `canvas` also. I'll aim for both.

Thanks for the feedback @mojavelinux. It's time to rebase the branch and start ticking off items in the TODO and get this merged.

obilodeau:
Rebase is done. Tasks are up-to-date. I'll be writing the docs in the coming days/weeks.

I don't think we should be concerned about the `data-background` backwards compatibility breakage at this point. The day we will tag some version then we should start worrying about it. What do you guys think?

mojavelinux:
:+1: 

> I don't think we should be concerned about the data-background backwards compatibility breakage at this point.

Agreed.

nicolaiparlog:
I like `background` because

1. as a non-Web-Developer it is much more obvious to me what it means (compared to `canvas`)
2. I appreciate the symmetry to reveal.js' `data-background` (a symmetry shared by _all_ attributes)

Maybe have both?

obilodeau:
Both: that's the plan

templates/slim/section.html.slim
```diff
@@ -1,6 +1,9 @@
 / hide slides on %conceal, %notitle and named "!"
 - titleless = (title = self.title) == '!'
 - hide_title = (titleless || (option? :notitle) || (option? :conceal))
+
+/ slide options processing
+- data_background = (attr? 'background') ? (image_uri(attr 'background')) : nil
```
kubamarchwicki: What about non image backgrounds? For reveal.js we have an option which doesn't seem supported
```html
<section data-background="#ff0000">
    <h2>All CSS color formats are supported, like rgba() or hsl().</h2>
</section>
```

```asciidoc
[background="#eee"]
== What are we doing?
```

is currently rendered into
```html
<section data-background="images/#eee">
   <h2>What are we doing?</h2>
</section>
```

mojavelinux: Great point. In my old patch, I had the following:

```
  - if (bg = attr 'data-background') && bg =~ /\.(bmp|gif|jpe?g|png|svg)$/
    - bg_image = image_uri bg
    - bg = nil
  - else
    - bg_image = (attr? 'data-background-image') ? image_uri(attr 'data-background-image') : nil
```

obilodeau: Agreed. This must definitely go in. I use background colors all the time.
How could I forget...

kubamarchwicki: Additionally (as I'm testing my old'ish slide decks with this PR), I was able to setup slides as follows

```asciidoc
[data-background="#eee", data-background-image=images/done-part1.gif, data-background-size=cover]
== No more Word design docs!
```

which resulted with a `section` having a class `has-light-background`. As per reveal.js release notes (version 3.0.0)

> The .reveal element is given a "has-light-background" or "has-dark-background" class when a per-slide background color is detected

Do you have any idea how it's possible now. I'm unable to set up a `class` attribute on `section` element, but having a `<section class="has-light-background">` will do the trick

kubamarchwicki:
+1 - would like to see that merged. Would be able to drop my hacky personal backend with the background images implementation.

kubamarchwicki:
As for the syntax, can we make it exactly the same as for asciidoctor-bespoke.js Dan is working on? Slides wise I think there will be two engines: reveal and bespoke and making the slides documents as universal as possible asciidoc wise seems like a tempting option

mojavelinux:
I think we agreed on the compatible syntax as:

```
image::cover.jpg[canvas,1024,640,size=cover,alt=Background image]
```

The reveal.js converter will also permit "background" in place of "canvas".

I haven't yet implemented this in the Bespoke.js converter.

mojavelinux:
The Bespoke.js converter will also support all the background attributes on the section as soon as that integration is complete. In other words, I'll track reveal.js and try to keep it aligned.

mojavelinux:
At the same time, we should recognize that we are still in the design phase and there will be differences between the converters as we explore new ideas. Eventually, everything will come into alignment as we learn the best way to write the AsciiDoc. I just want to make sure we don't tie our hands down too early.

kubamarchwicki:
On 30 Mar 2016 10:24, "Dan Allen" <notifications@github.com> wrote:
>
> I think we agreed on the compatible syntax as:
>
> image::cover.jpg[canvas,1024,640,size=cover,alt=Background image]
>
> The reveal.js converter will also permit "background" in place of
"canvas".
>

Great! I'll catchup with the docs and examples to get this one merge in.
Least I can do :)

Kuba

obilodeau:
This is ready to go in. Can someone acknowledge it please? There are test cases to be rendered in `test/` you can use `make` to run asciidoctor on all of them.

templates/slim/section.html.slim
```diff
@@ -1,12 +1,48 @@
...
+  - data_background_color = attr 'background-color'
+
+/ RENDERING
 / render parent section of vertical slides set
 - if @level == 1 && !vertical_slides.empty?
   section
-    section id=(titleless ? nil : @id) data-transition=(attr 'data-transition') data-transition-speed=(attr 'data-transition-speed') data-background=(attr 'data-background') data-background-size=(attr 'data-background-size') data-background-repeat=(attr 'data-background-repeat') data-background-transition=(attr 'data-background-transition')
+    / TODO: try to get rid of duplication w/ standalone slide section
+    section(id=(titleless ? nil : @id)
+      class=[@style,role]
```
mojavelinux: I don't think `@style` is really needed there. The style of a section is mostly reserved for special sections in DocBook. You can simply write this as:

```
class=roles
```

templates/slim/block_image.html.slim
```diff
@@ -1,10 +1,11 @@
-.imageblock(id=@id class=[@style,role]
-    style=[("text-align: #{attr :align}" if attr? :align),("float: #{attr :float}" if attr? :float)].compact.join('; '))
-  .content
-    - if attr? :link
-      a.image href=(attr :link)
+- unless attr(1) == 'background' || attr(1) == "canvas"
+  .imageblock(id=@id class=[@style,role]
```
mojavelinux: I don't think an image block can have a style.

obilodeau: I have to admit, I don't know what `@style` does and where it comes from... I just copied and pasted stuff around. I'm a sham. ;)

templates/slim/section.html.slim
obilodeau: Fixed

templates/slim/block_image.html.slim
obilodeau: Fixed. I kept `class=roles`, I hope that's fine.

mojavelinux: Yep, that's good. That will propagate all roles defined on the block, e.g.

```
[.canvas]
image::image.jpg[]
```

test/data-background-newstyle.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+image::meme-7.png[]
+
+== !
+
+image::70s.jpg[background,size=contain]
+
+== hey
+
+image::bio.jpg[background,size="100px"]
```
mojavelinux: Technically, the quotes around an attribute value are only required if the value has commas or spaces. Just FYI.

templates/slim/block_image.html.slim
```diff
@@ -1,10 +1,11 @@
-.imageblock(id=@id class=[@style,role]
-    style=[("text-align: #{attr :align}" if attr? :align),("float: #{attr :float}" if attr? :float)].compact.join('; '))
-  .content
-    - if attr? :link
-      a.image href=(attr :link)
+- unless attr(1) == 'background' || attr(1) == "canvas"
```
mojavelinux: I tend to prefer to use single quotes unless interpolation is used in the string. (e.g., `'canvas'`)

README.adoc
```diff
@@ -63,20 +63,89 @@ Actually things aren't that bad
...
+It sets `reveal.js'` `data-background-image` attribute.
+The `size` attribute is also supported.
+See the https://github.com/hakimel/reveal.js/#image-backgrounds[
+reveal.js' documentation] for details.
+
+NOTE: Background images file names are now relative to the `:imagedir:`
+attribute if set.
+
+NOTE: The `canvas` keyword can be used instead of `background` for the same
+effect.
```
mojavelinux: I prefer to use the sentence-per-line formatting instead of hard wrapping at a fixed column. Would you agree to use that style?

mojavelinux:
I added some inline comments. Other than that, I think we're good to go.

mojavelinux:
There's one thing I want to point out, but its not specific to this change. The `attr` and `attr?` methods inherit by default. That means if it doesn't find the attribute name on the node, it looks on the document. Generally, you only want to enable this behavior if you intend to allow a setting to be controlled globally. That might be good in the case of transition, for instance:

```
= My Slides
:transition-speed: fast

== First Slide
```

However, there may be attributes that you don't want to inherit. If that's the case, you generally use the form:

```
attr('name', nil, false)
```

The second parameter value is the default attribute value, which is `nil` by default.

README.adoc
obilodeau: I'm doing sentence-per-line with a hard-wrap at 80. I assumed it was ok like that. I'll fix it.

templates/slim/block_image.html.slim
obilodeau: Python spoiled me on this. I'm a former Perl dev and I understand totally the right quoting for the right context.

obilodeau:
I added the attribute inheritance note to `HACKING.adoc` as I am afraid I'll forget. Also did the cleanup. I'm merging this. Thanks!

mojavelinux:
:+1:

Issue #30, opened 2015-05-09
Footnotes don't appear


wackywendell:
Footnotes do not seem to appear:

``` asciidoc
= A Test Document
Author Name
:revealjs_theme: simple
:revealjs_transition: none
:revealjsdir: https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.0.0/
:stem: latexmath
:imagesdir: imgs

== Test

This is some text with footnote:[An example footnote.]
```

The text of the footnote doesn't seem to show up in the HTML:

``` html
<section id="_test"><h2>Test</h2><div class="paragraph">
  <p>This is some text with 
    <span class="footnote">[<a id="_footnoteref_1" class="footnote" href="#_footnote_1" title="View footnote.">1</a>]</span>
  </p>
</div></section>
```

There's more HTML than that, of course, but nothing with the footnote text in it.

mojavelinux:
Footnotes are not implemented in the reveal.js backend. But before we add them, we need to make a design decision. Where should the footnotes appear? Should they appear at the bottom of the slide from which they come? Should they appear on a references slide at the end of the presentation?

It is possible to check for new footnotes at the bottom of a slide and add them. The API is a little clunky, but I believe the necessary information is there.

wackywendell:
Good questions / points. I assumed "footnote" would mean "bottom of slide", given that it normally means "bottom of page", but that is a design assumption...

mojavelinux:
Let's start with the bottom of the slide.

keithbuhler:
I'll add my vote to putting them on the bottom of the slide. For now, I'm just hyperlinking to references.

obilodeau:
Another option to consider: https://github.com/e-gor/Reveal.js-Title-Footer

lalebarde:
With Asciidoctor PDF 1.5.0.beta.1 using Asciidoctor 2.0.10, it goes to the last slide. Why not distinguishing:

- footnote : in the bottom of the slide

- endnote : in a slide declared as 'endnotes' receipient (not necessarily the last one)

Ezwen:
This would be great to have! I tend to put references to other papers/works at the bottom of slides using footnotes.

obilodeau:
I'll explore this for the next release

Mogztter:
> Let's start with the bottom of the slide.

In this case, I think we should probably reset the numbering on each slide right?

| | |
| -- | -- |
[...]

PR #53, opened 2015-10-22
Footers for Reveal.js Slides

ysb33r:
Instead of pushing the footer rigt at the end as is the case with normal HTML pages, we sneak it in just at the begining of the Reveal block at the lock the div to the bottom of the page. This allows for the docinfo.html and docinfo-footer.html to be used to easily insert a footer.

All you need to do is to add some content in to the `docinfo-footer.html` file i.e.

```
Use arrows to navigate left and right
```

and at some CSS in `docinto.html` that operates on `div.footer` i.e.

```
<style>
    div.footer {
        font-size: 0.5em;
    }
</style>
```

If `:docinfo`:` is set the asciidoc files, then the footer will appear on each slide.

_Caveats_:  Curretnly the footer is not yet added to the print-pdf functionality. However once people look at this, someone else might figure it out before me how to complete that step.

### DIFF
--- a/templates/slim/document.html.slim
+++ b/templates/slim/document.html.slim
@@ -69,6 +69,9 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
       link rel='stylesheet' href=((customcss = attr :customcss).empty? ? 'asciidoctor-revealjs.css' : customcss)
   body
     .reveal
+      - unless (docinfo_content = (docinfo :footer, '.html')).empty?
+        .footer style='position: absolute; bottom: 1em; left: 1em'
+          =docinfo_content
       / Any section element inside of this container is displayed as a slide
       .slides
         - unless notitle || !has_header?
@@ -157,7 +160,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
         parallaxBackgroundImage: '#{attr 'revealjs_parallaxbackgroundimage', ''}',
         // Parallax background size in CSS syntax (e.g., "2100px 900px")
         parallaxBackgroundSize: '#{attr 'revealjs_parallaxbackgroundsize', ''}',
-        
+
         // The "normal" size of the presentation, aspect ratio will be preserved
         // when the presentation is scaled to fit different resolutions. Can be
         // specified using percentage units.
@@ -170,7 +173,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
         // Bounds for smallest/largest possible scale to apply to content
         minScale: #{attr 'revealjs_minscale', 0.2},
         maxScale: #{attr 'revealjs_maxscale', 1.5},
-        
+
         // Optional libraries used to extend on reveal.js
         dependencies: [
             { src: '#{revealjsdir}/lib/js/classList.js', condition: function() { return !document.body.classList; } },
@@ -181,5 +184,3 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
             { src: '#{revealjsdir}/plugin/notes/notes.js', async: true, condition: function() { return !!document.body.classList; } }
         ]
       });
-    - unless (docinfo_content = (docinfo :footer, '.html')).empty?
-      =docinfo_content
--- a/templates/slim/document.html.slim
+++ b/templates/slim/document.html.slim
@@ -68,10 +68,10 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     - if attr? :customcss
       link rel='stylesheet' href=((customcss = attr :customcss).empty? ? 'asciidoctor-revealjs.css' : customcss)
   body
+    - unless (docinfo_content = (docinfo :footer, '.html')).empty?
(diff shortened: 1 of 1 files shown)


mojavelinux:
If inserting the footer HTML at this location is necessary to get the overlay on slides to work, I support it.

While reviewing this pull request, I noticed that the `:header` document argument for the header docinfo should be removed so that it works in both 1.5.2 and 1.5.3.

```
- unless (docinfo_content = docinfo).empty?
```

The reason is that in 1.5.3, the option name changed to `:head`...but it's still the default, so just not specifying it allows it to work despite the change.

mojavelinux:
cc @obilodeau thoughts?

obilodeau:
It's kind of sad that upstream reveal.js doesn't have footers more integrated and that we have to make up our own standard... I've never used docinfo files with asciidoctor yet and from what I recall from asciidoc, asciidoctor seems to be doing things differently (ex: I don't recall having a docinfo.html file).

In all honesty, I would love to see an example source that I could render on my own. @ysb33r: if you could provide a minimal example in a [gist](https://gist.github.com/) or its own repo that would be best. It would make it easier to evaluate your proposal.

Also, it would need to be documented.

ysb33r:
@obilodeau, All you need to do is to have a `docinfo-footer.html` file in same root directory as the base document. I've just built on the docinfo convention that is already in use for html/docbook backends.

As an example, this is the  `docinfo-footer.html` I used for Devoxx Morocoo.

```html
<img src="images/devoxxma.png" style="height : 183px; width : 50px"/>
<table style="width : 1200px; background-color : #D96625; color : white; box-shadow: 3px 3px 3px rgb(254,193,0);">
    <tr>
        <td style="width : 25%; text-align : left">#DevoxxMA</td>
        <td style="width : 50%; text-align : center">@DevoxxMA</td>
        <td style="width : 25%; text-align : right">© Schalk W. Cronjé 2015</td>
    </tr>
</table>
```
which renders as scan be seen on the first slide of http://www.slideshare.net/ysb33r/idiomatic-gradle-plugin-writing-55261739. *Note* that this renders differently for the print view as for the screen view. In screen view the footer is on all pages. In print view the footer is only on the first page. I hve not had the tie to figure out what needs to be done to have screen & print view the same.

mojavelinux:
> asciidoctor seems to be doing things differently (ex: I don't recall having a docinfo.html file).

That's the same from AsciiDoc Python. What Asciidoctor adds is a second insertion point for docinfo at the document footer.

Btw, I don't we have to add the footer content above the slides. Using CSS, it should be possible to move the z-index of the footer content up so that they overlay on the slide canvas.

ysb33r:
I'm all up for a better solution. I'm no CSS expert and this was my first real Slim hack too. I would just like to have something that people can use, even if it means merge now and make better later.

mojavelinux:
Understood.

It's very likely that the print-pdf just requires some additional styles to make the footer part of the page view. I've worked with the print styles in reveal.js before, so likely I can sort out what needs to get added. But I agree that shouldn't hold up the merge.

obilodeau:
This addresses my concerns. All that is missing for a first merge is that
you document the feature in the README with a short example.

mojavelinux:
:+1:

obilodeau:
@ysb33r still interested in providing the documentation?

obilodeau:
Just realized this relates to #30.

ysb33r:
My apologies. These docs have been a long time outstanding.

obilodeau:
@ysb33r can you allow us to push into the branch? This way we could help you with the documentation. [Here is how to do it](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/#enabling-repository-maintainer-permissions-on-existing-pull-requests).

obilodeau:
Here is another approach to footers: https://github.com/CodeFX-org/slides/blob/master/_shared/event-on-title-and-footer.adoc (as referred to in #123). I'll try to evaluate both and see which one integrates best.

ehmkah:
@obilodeau  - did you make a decision which solution for the footer you are going to integrate? Would be cool to add a visible footer to the presentation...

obilodeau:
I haven't. I'm busy with some infrastructure changes (doctest and compiling slim templates into ruby) at the moment but I'll evaluate it eventually. These changes will help tests contributions like this better.

Providing an example and documentation here or in the PR of how to specify a footer in AsciiDoc would accelerate that evaluation because right now the documentation of @ysb33r's proposal still doesn't exist.

obilodeau:
Another option to consider: https://github.com/e-gor/Reveal.js-Title-Footer

bcouetil:
I'm using the mentionned Reveal.js plugins with success. I guess we don't need some custom asciidoctor-reveal.js code anymore 😄

ysb33r:
I am killing this PR because of what @bcouetil said.

mojavelinux:
That's actually a good thing. The more pure reveal.js can be, the greater the flexibility.

PR #64, opened 2016-03-29
Small remarks about gems

kubamarchwicki:
The latest asciidoctor-reveal.js backend didn't work with asciidoctor-1.5.3 and titl 2.0.1 - which i used to have installed

This is a small but pesky one related to this exact comment: https://github.com/asciidoctor/asciidoctor-reveal.js/pull/52/commits/83a1097086516d6e84cad6abab3bfcb47f3358e2#r56757487 :)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -9,7 +9,7 @@ The https://github.com/asciidoctor/asciidoctor-reveal.js[reveal.js backend] is a
 
 == Prerequisites
 
-. Ensure Asciidoctor, Slim and their dependencies are installed:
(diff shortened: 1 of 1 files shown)


mojavelinux:
I think situations like this can be resolved once we start having releases. That at least allows you to move between releases easily and we can communicate breaking changes via release numbers.

I've demonstrated how to make a gem for the template-based converters in asciidoctor-bespoke.

mojavelinux:
The PR no longer merges. I don't understand why though because it doesn't look like there's a merge conflict.

obilodeau:
This can be closed due to a minimum version check implemented in 81e1139.

Issue #68, opened 2016-04-26
Enhancement : Content spanning on multiple slides


GrmpCerber:
I've got a document which content is suitable for both PDF and reveal presentation.
Yet, as pointed out by #59 title level 3 and below are ignored, thus ending in content being unreadable.

In addition to #59 I would say that "level" based splitting is not enough: as an example, I’ve got a section that holds a list and two schemas. 
Only the list part is visible. 
I would like to be able to show the two schema on different slides without altering the PDF display (thus not using `<<<` as slide splitter)

obilodeau:
What version of asciidoctor-reveal.js are you using? Current master has this in https://github.com/asciidoctor/asciidoctor-reveal.js/commit/f027d08cd22a5e1ca27a46b96dd380abe062ba0a which means that level 3 and below are no longer ignored and should render in your PDF.

As for your second remark, can you give a short example in AsciiDoc? I don't understand.

GrmpCerber:
I've discovered one could create a fake section by adding 
    
    === !
    
(Preceded and followed by a blank line)
Yet this is not practical as it impact the PDF document ...

@obilodeau I did a shallow clone yesterday so my version is d39f64d + Asciidoctor 1.5.4

As for a short example I did so in the attached file 
[test.adoc.txt](https://github.com/asciidoctor/asciidoctor-reveal.js/files/236804/test.adoc.txt) and compiled it with : 

    asciidoctor -r asciidoctor-diagram -T ./asciidoctor-reveal.js/templates/slim/ test.adoc

My vertical slide visually ends with the level 3 title.
As you could guess my example could be visually splited in 4 slides : 
* level 2 title and explaination 
* level 3 title with list 
* and then the two schemas on two slides

mojavelinux:
> I've discovered one could create a fake section..., Yet this is not practical as it impact the PDF document ...

Keep in mind you can use preprocessor conditionals to hide these fake sections from other backends:

```
ifndef::backend-pdf[=== !]
```

You can also do all sorts of other things using preprocessor conditionals.

GrmpCerber:
Nice !
This should be added to the readme (you or I)

This said, it's not 100% practical since `=== !` are specific to the reaveal.js backend, hence I would suggest adding a `backend-revealjs` attribute.
In addition It would also be a nice feature to be able to repeat the last title (with `=== !repeat` for example)

mojavelinux:
[...]

PR #69, opened 2016-04-27
Update README.adoc

GrmpCerber:
Fixing asciidoctor/asciidoctor-reveal.js#68

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -179,6 +179,35 @@ include::test/concealed-slide-titles.adoc[]
 
 NOTE: `conceal` and `notitle` have the advantage that the slide still has an id so it can be linked to.
 
+=== Splitting long slides without impacting other backends
+To mark slides split points you can use preprocessor conditionals combinded with backend declaration.+
+Breaking points are set with slideless titles `=== !` wrapped in a conditionnal : `ifdef::backend-revealjs[=== !]`. 
+In the end, the whole document has to be compiled with the backend option : `-b revealjs`
+
+For example :
+[example]
+----
+== Main section
+
+=== Sub Section    
+Small +
+Multiline +
+intro
+
+. very
+. long
+. list
+. of
+. items
+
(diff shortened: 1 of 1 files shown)


obilodeau:
I tried to reproduce your documentation with the adoc file you provided in #68 and I don't arrive at your results.

The `ifdef::backend-revealjs[=== !]` is visually rendered inline both in with the reveal.js back-end and the html back-end.

```
$ asciidoctor -V
Asciidoctor 1.5.4 [http://asciidoctor.org]
Runtime Environment (ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]) (lc:UTF-8 fs:UTF-8 in:- ex:UTF-8)
```

I merged your PR in this branch and added my test case: https://github.com/asciidoctor/asciidoctor-reveal.js/compare/content-for-multiple-backends#diff-eb6036f05534957c98e164f1a6f20250R1

Can you tell me what I'm missing? Thanks!

GrmpCerber:
I'm not 100% sure I understood your question correctly but I think I spotted an error in your test case (see my PR #80 )
PR #69 content was meant to be displayed by github.com as an hint on how to redact the adoc source, but you merged the raw content with my test file to create  your test case and ... BAM ! :)
In `ifndef::escapingSillyPreprocessor[ifdef::backend-revealjs[=== !]]` "escapingSillyPreprocessor" was meaning : github.com please stop trying to parse this because I want "ifdef::backend-revealjs[=== !]" to be displayed as is to the visitor

obilodeau:
Got it! It works. Thanks!

Although your solution is very clever, you can escape a directive with a backslash:

```
\ifdef::backend-revealjs[=== !]
```

See http://asciidoctor.org/docs/user-manual/#preventing-substitutions

GrmpCerber:
@obilodeau Sorry to have missed that :) `\` indeed seems easier to understand

Issue #32, opened 2015-05-13
section id starting with underscore breaks starting from specific slide in url


bartoleo:
reveal.js on page open checks if url contains a slide 'id' as hash (function readURL())
sections generated by asciidoctor-reveal have id starting with underscore 
reveal.js checks if the id found is valid
```
// Ensure the named link is a valid HTML ID attribute
			if( /^[a-zA-Z][\w:.-]*$/.test( name ) ) {
```
and it's always Not valid so it starts from beginning

I made a workaround in /asciidoctor-reveal.js/templates/slim/section.html.slim, adding 'slide' before id:
 'slide'+@id

mojavelinux:
I've noticed this as well.

I think this simply needs to be documented. You must set the `idprefix` in your AsciiDoc document.

```
= Presentation Title
:idprefix:
```

or

```
= Presentation Title
:idprefix: slide_
```

We could see if it's possible to set a better default from the converter to save you the trouble.

bartoleo:
perfect, thank you!!!

I have still a problem with id containing accented letters like à è ì ò ù (and uppercase version)
reveal.js during url writing strips these characters

I made another workaround in /asciidoctor-reveal.js/templates/slim/section.html.slim, replacing characters in id:
```
@id.gsub(/[^a-zA-Z0-9\-\_\:\.]/, '')
```

mojavelinux:
That's a reasonable workaround to consider adding to this backend.

I think the real problem here is the logic in reveal.js. Asciidoctor purposefully permits all characters that are allowed by the HTML spec. reveal.js is being overly conservative by stripping these. `à` is a perfectly valid letter character and reveal.js is showing its bias towards English by stripping it.

Having said that, I'm still prepared to do what we have to do to make it work (esp if reveal.js refuses to change). Keep in mind that we'll need to think about references to those ids (i.e., inline_anchor), though in a presentation that's less critical.

bartoleo:
I'll open an Issue on reveal.js asking to remove this constraints

mojavelinux:
Thanks!

bartoleo:
opened:
https://github.com/hakimel/reveal.js/issues/1230

obilodeau:
Fixed. See #75.

mojavelinux:
@obilodeau Can you add a comment in the source code that the workaround is temporary. In the long run, we don't want to be manipulating the ID that Asciidoctor generates as it can break xrefs.

PR #75, opened 2016-07-14
Limit named links to [a-zA-Z0-9\-\_\:\.]

kubamarchwicki:
When working with other languages than English there are often nonlatin characters in section titles. When reveal.js is working with history enabled - this doesn't get resolved to proper navigation links (presentation starts over from the begining)

This is due to this reveal.js issue https://github.com/hakimel/reveal.js/issues/836 which resulted in a following pull request in reveal.js itself https://github.com/hakimel/reveal.js/commit/da82c8ce81a5acc8b14002a5818a32ce0c89f7b8

The important change is 
```
id = id.toLowerCase();
id = id.replace( /[^a-zA-Z0-9\-\_\:\.]/g, '' );
```

which I've added to the section template as well. 

I'm not sure if that's the way it should be done so feedback is appreciated

### DIFF
--- a/templates/slim/section.html.slim
+++ b/templates/slim/section.html.slim
@@ -2,6 +2,7 @@
 - titleless = (title = self.title) == '!'
 - hide_title = (titleless || (option? :notitle) || (option? :conceal))
 - vertical_slides = find_by(context: :section) {|section| section.level == 2 }
+- @id = @id.gsub(/[^a-zA-Z0-9\-\_\:\.]/, '')
 
 / render parent section of vertical slides set
 - if @level == 1 && !vertical_slides.empty?


mojavelinux:
The question is whether this should be changed in the converter or upstream in reveal.js. I'd like to see it fixed in reveal.js.

Having said that, if we do decide to offer a patch in the interim, it would be better to modify the id at the point it is generated. That would mean overloading the generate_id method on the Section class. That way, the id is right in the model.

kubamarchwicki:
My original intention was to have those two (revealjs and backend) follow same rules (adhere to same expectations). I think it currently drills down to a fact neither revealjs not backend is implementation for url fragments. 

As the backend takes the section title as the URL fragment - any character is allowed. This is not right from the RFC standpoint (https://tools.ietf.org/html/rfc3986#section-3.5, which was nicely TL;DR'ed [on stackoverflow](http://stackoverflow.com/questions/26088849/url-fragment-allowed-characters/26119120#26119120))
Bottom line is revealjs removes too many characters, backend removes too little. I'd be happy to submit PR to have it fixed in both places

Back to asciidoctor, I totally agree the right place for id would be in the model. My hacking around was 'a foot in the door' to get things started. Can you please point me to an example of a backend that overloads model classes? That would be of a great help for me - I haven't been there yet :)

obilodeau:
@mojavelinux I tried to find an example of class overloading from `helper.rb` in a back-end and couldn't find anything. Can you point us to an example please?

obilodeau:
Related to #32

obilodeau:
Decided to merge. I added a few changes on top. See commit tagged with this PR # for details. Most importantly the regex anchor (^) was inside the character class and, in my testing, it didn't work.

mojavelinux:
> As the backend takes the section title as the URL fragment - any character is allowed.

The id / fragment is derived from the section title, but it is not equivalent to it. The id generator in Asciidoctor does take the RFC rules into account. See https://github.com/asciidoctor/asciidoctor/blob/v1.5.4/lib/asciidoctor/section.rb#L89

> Bottom line is revealjs removes too many characters, backend removes too little. 

The bottom line is that Reveal.js is out of compliance.

The problem with changing the ID is any xref (which contains non-ASCII characters) to that ID from within the content is no longer going to work. But there's no much we can do to avoid that until the ID matching in Reveal.js is fixed. With the current patch, we simply run the risk of breaking xrefs (which contain non-ASCII characters).

mojavelinux:
Upstream issues:

* https://github.com/hakimel/reveal.js/issues/1230
* https://github.com/hakimel/reveal.js/issues/1346

As a follow-up, I think we need to submit a patch to Reveal.js. I'd also like to see a comment in the code that ID manipulation is temporary.

Issue #82, opened 2016-08-28
Applying reveal.js' stretch class doesn't give intended result for images


obilodeau:
Revealjs has a feature called stretch to give the focus to a specific HTML element: https://github.com/hakimel/reveal.js#stretching-elements

It works by using javascript to assign a style with height and width to the specific element. The javascript is not executed when changing slides since I tried setting the attribute directly in chrome's developer console and it did not work.

One problem is the following limitation of the stretch feature:
> Only direct descendants of a slide section can be stretched

Currently, this back-end wraps in HTML tags heavily. Images are wrapped in `imageblock` then a `div.content` and finally the image tag (if there's no link). The nesting means `stretch` won't work.

We need to think about what we can do to benefit from this `stretch` class which applies directly the exact amount of pixels we can expand to without clobbering the rest of the content.

mojavelinux:
> Currently, this back-end wraps in HTML tags heavily. Images are wrapped in imageblock then a div.content and finally the image tag (if there's no link). The nesting means stretch won't work.

Don't hesitate to change the HTML structure. As I mentioned in another comment, I'd like to see modern HTML used for images like what's in the Bespoke converter (i.e., figure/figcaption). (See https://github.com/asciidoctor/asciidoctor-bespoke/blob/master/templates/slim/image.html.slim). In general, we should be gravitating towards modern HTML which also works best with Reveal.js. We have to strike a balance.

obilodeau:
Yes, let's aim to do modern HTML wherever we can.

That said, there is a conflict here. First is that using `figure` means wrapping `img` in another tag and will make `.stretch` not work (it works only to direct descendants of slide content). Second, as I said in #81, re-doing the equivalent of what .stretch does in pure CSS really challenges my CSS skills.

So I would close this since it's fixed in master. Then the day someone figures it out in pure CSS  (pun unintended) we'll go down the `figure` road.

Thoughts?

obilodeau:
No comments, closing. Open a new bug if necessary.

mojavelinux:
> So I would close this since it's fixed in master. Then the day someone figures it out in pure CSS (pun unintended) we'll go down the figure road.

[...]

PR #83, opened 2016-08-28
Changes to block image behavior to support .stretch (#82)

obilodeau:
If height or width is unspecified and the stretch class is set, we will
stretch the height of the image to 100%. This enables us to keep proper
aspect ratio and fill available slide space.

To do so we had to remove the intermediary div.content.


templates/slim/block_image.html.slim
```diff
@@ -1,10 +1,21 @@
...
+
+/ When the stretch class is present, block images will take the most space
+/ they can take. Setting width and height can override that.
+/ We pinned the 100% to height to avoid aspect ratio breakage and since
+/ widescreen monitors are the most popular, chances are that height will
+/ be the biggest constraint
+- if (has_role? 'stretch') && !((attr? :width) || (attr? :height))
+  - height = "100%"
+
+- unless attr(1) == 'background' || attr(1) == "canvas"
```
obilodeau: This is a change that is in #52 that I forgot to remove on a merge conflict

mojavelinux:
> To do so we had to remove the intermediary div.content.

Yes, please feel free to customize the HTML to complement Reveal.js. Keep in mind that the default HTML output in Asciidoctor is legacy HTML and is by no means the standard. You'll see in Bespoke i changed the HTML quite dramatically. For instance, I wrap the `<figure>` element around image elements. Do what works best...as long as you honor the settings from the source document such as the roles.

templates/slim/block_image.html.slim
```diff
@@ -1,10 +1,21 @@
...
+  .imageblock(id=@id class=[@style,role]
```
mojavelinux: I'd like to see modern HTML used for images like what's in the Bespoke converter (i.e., figure/figcaption). See https://github.com/asciidoctor/asciidoctor-bespoke/blob/master/templates/slim/image.html.slim In general, we should be gravitating towards modern HTML which also works best with Reveal.js. We have to strike a balance.

Issue #63, opened 2016-03-20
Jade templates are behind slim templates


obilodeau:
Bugs in jade templates:
- [x] block_thematic_break template is missing (fixed in #90)
- [x] highlightjs issue (#88)

Small missing features in the jade templates:
- [ ] feature: add `revealjsdir` attribute support to control the location of reveal.js (make sure it works both on local filesystem paths (cross-platform) and URL (ex: https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.3.0/)
- [ ] feature: speaker notes
- [ ] feature: vertical slides
- [ ] feature: concealed titles
- [ ] feature: :customcss: (#85)

Other misc things:
- [ ]  [jade was renamed pug](https://github.com/jadejs/jade) (#102)
- [ ] have a way to quickly test templates changes ([current method](https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/HACKING.adoc#jade-templates)) 
- [ ] integrate rendering of the tests in the `Makefile`
- [ ] document.jade: Support attr - asset_uri_scheme, revealjs_customtheme, iconfont-name
- [ ] document.jade: Slim reference FontAwesome 4.3 while jade 4.4.

mojavelinux:
>  jade was renamed pug

I'm confused when this is supposed to take effect. I guess the first official release of pug is still pending. However, since this is new code, perhaps we should start targeting pug now. (Though, the transition from jade to pug is likely going to take a long time).

rahmanusta:
It is frustrating to see "rename" on a known project.

@obilodeau Thanks for the missing parts. We should update these parts.

mojavelinux:
> It is frustrating to see "rename" on a known project.

100% agree. In this case, though, it seems they were forced into a legal corner and had to change it to avoid a costly battle. I'm not sure about the name pug though. Very strange.

cmoulliard:
I have discovered another jade/slim file which is missing 

```
asciidoctor-core.js:880 GET http://localhost:8080/build/examples/templates/revealjs/thematic_break.jade 404 (Not Found)
jade.js:3831 Uncaught IOError: No such file or directory: templates/revealjs/thematic_break.jade on line 8 on line 25
```

This jade file does not exist & corresponds to https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb#L751

It has been implemented within the asciidoctor-backend project 

[...]

PR #85, opened 2016-08-29
:customcss: attribute for easy per-presentation CSS

obilodeau:
Check the documentation commit for details.

### DIFF
--- a/templates/slim/document.html.slim
+++ b/templates/slim/document.html.slim
@@ -65,10 +65,8 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
       document.write( '<link rel="stylesheet" href="#{revealjsdir}/css/print/' + ( window.location.search.match( /print-pdf/gi ) ? 'pdf' : 'paper' ) + '.css" type="text/css" media="print">' );
     - unless (docinfo_content = docinfo :header, '.html').empty?
       =docinfo_content
-    - if attr? :customcss and (attr :customcss).empty?
-      link rel='stylesheet' href='asciidoctor-revealjs.css'
-    - elsif attr? :customcss
-      link rel='stylesheet' href='#{attr :customcss}'
+    - if attr? :customcss
+      link rel='stylesheet' href=((customcss = attr :customcss).empty? ? 'asciidoctor-revealjs.css' : customcss)
   body
     .reveal
       / Any section element inside of this container is displayed as a slide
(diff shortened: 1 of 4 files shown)


mojavelinux:
I like the capability, but we're changing the meaning of the linkcss attribute attribute here. Therefore, I think we should use a different attribute name. We could align with the deck.js converter and use the customcss attribute instead (or perhaps revealjs_customcss). If empty, it resolved to asciidoctor-reveal.css. If non-empty, the value provided is used.

obilodeau:
Sounds fair. I'll go with `customcss` since this is not related to an upstream feature.

mojavelinux:
:+1:

cmoulliard:
Don't forget to also update the jade template !

obilodeau:
@cmoulliard as I said in #63, I won't do the jade templates. I love asciidoc[tor] and I love sharing about infosec with nice HTML presentations but I don't have enough time to learn properly ruby and javascript plus slim and jade/pug. I already feel my contributions are not [idiomatic](http://mrjoelkemp.com/2013/05/what-is-idiomatic-programming/) to a ruby programmer (being a Python/Perl "programmer") and I don't want to double that mess. Someone needs to step in to do the jade thing or we need to switch to a ruby/javascript accessible template engine.

obilodeau:
@mojavelinux I implemented what we discussed and added a test case. From my perspective, this is ready to go.

templates/slim/document.html.slim
```diff
@@ -65,6 +65,10 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
       document.write( '<link rel="stylesheet" href="#{revealjsdir}/css/print/' + ( window.location.search.match( /print-pdf/gi ) ? 'pdf' : 'paper' ) + '.css" type="text/css" media="print">' );
     - unless (docinfo_content = docinfo :header, '.html').empty?
       =docinfo_content
+    - if attr? :customcss and (attr :customcss).empty?
```
mojavelinux: Here's my preferred way of writing this:

```slim
- if attr? :customcss
  link rel='stylesheet' href=((customcss = attr :customcss).empty? ? 'asciidoctor-revealjs.css' : customcss)
```

I'd also be open to:

```slim
- if attr? :customcss
  - if (customcss = attr :customcss).empty?
    link rel='stylesheet' href='asciidoctor-revealjs.css'
  - else
    link rel='stylesheet' href=customcss
```

mojavelinux:
> I already feel my contributions are not idiomatic to a ruby programmer (being a Python/Perl "programmer")

I think you're doing a fine job. We're learning together. Besides, we use a unique Ruby style in this project anyway, and I kind of like that.

cmoulliard:
Even if you can't maintain the templates in jade, it is important to list somewhere the changes applied to slim and/or jade.

templates/slim/document.html.slim
obilodeau: I noticed that you love the ternary operator :)

obilodeau: The test/assignment is very interesting, I'll try to remember that trick!

mojavelinux: I love it to death. It helps cut down on a lot of lines of code and useless assignments. Combined with the assignment in the conditional, I can express meaning much more succinctly.

PR #87, opened 2016-09-02
Formalized a merge / review policy

obilodeau:
Open for comments. Comes from an email discussion between me and @mojavelinux.

### DIFF
--- a/HACKING.adoc
+++ b/HACKING.adoc
@@ -152,3 +152,19 @@ If that's the case, you generally use the form:
 The second parameter value is the default attribute value, which is nil by default.
 
 Relevant documentation: http://www.rubydoc.info/github/asciidoctor/asciidoctor/Asciidoctor%2FAbstractNode%3Aattr
+
+
+== Merge / Review policy
+
+Any non-trivial change should be integrated in master via a pull-request.
+This gives the community a chance to participate and helps write better code because it encourages people to review their own patches.
+
+Once a pull request is submitted, let it sit for a minimum of 24 hours for small changes.
+Larger changes should sit longer at around a week.
+Breaking changes should sit until a prominent contributor comments on the changes.
+Ping `@mojavelinux` and `@obilodeau` if necessary.
+
+If you don't get feedback, just merge after the sitting time frame.
+
+Remember that this is a slower moving project since people are not designing slides everyday.
+Well, for most people.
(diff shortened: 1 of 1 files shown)


mojavelinux:
Looks good to me!

The only other think I would add is that all pull requests should come from a fork. It's best not to clutter the upstream with a bundle of development branches.

As for the 24 hour rule, I'd also add that if you get a bunch of :+1: in a short period, then that's another trigger to proceed. It need not just be time based. I'd also say 24-48 hours instead of just 24.

obilodeau:
> The only other think I would add is that all pull requests should come from a fork. It's best not to clutter the upstream with a bundle of development branches.

Does it really matter? Upstream branches on the other hand have the ability to be modified by every team member. I've forked a fork several times in the past, especially to finish or modify documentation. Working on an upstream branch would have made the whole thing cleaner.

obilodeau:
> As for the 24 hour rule, I'd also add that if you get a bunch of :+1: in a short period, then that's another trigger to proceed. It need not just be time based. I'd also say 24-48 hours instead of just 24.

Just committed changes in 32ad081.

mojavelinux:
> Does it really matter? Upstream branches on the other hand have the ability to be modified by every team member.

I'm pretty adamant about this one. I really don't like to see "workspace" branches in the upstream. They get too easily mixed up with maintenance branches...or just get stale and left behind.

While it's true that this prevents team members from pushing to the same branch, I actually think that's a good thing. It continues to promote the propose and review model throughout the workflow.

In the end, it's up to you what you choose for this project. But I feel strongly that keeping branches in the forks keeps everything more clear and organized.

mojavelinux:
I forgot to mention that GitHub now allows project maintainers / committers to push directly to the branch of an open pull request. That promotes the workflow of issuing pull requests from the branch of a fork.

obilodeau:
> I forgot to mention that GitHub now allows project maintainers / committers to push directly to the branch of an open pull request. That promotes the workflow of issuing pull requests from the branch of a fork.

I missed that. Cool feature!

I just added a sentence about the fact that PRs should come from forks: 813e823. Also, I'll change my behavior starting now so my devel branches should gradually disappear.

mojavelinux:
Awesome! If you find the fork-based workflow makes it too difficult to maintain the project, I'm happy to revisit. But I that appreciate you're willing to give it a shot. :beers:

Issue #78, opened 2016-08-26
Turn asciidoctor-reveal.js into a proper back-end


obilodeau:
Need to add code like this to the repository: https://github.com/asciidoctor/asciidoctor-bespoke/blob/master/lib/asciidoctor-bespoke/converter.rb

As discussed in #68

mojavelinux:
See sample code in the following comment:

https://github.com/asciidoctor/asciidoctor-reveal.js/issues/68#issuecomment-243336091

obilodeau:
Thanks for that. Looking forward to implement that change. I looked at the bespoke back-end a bit yesterday and I'm really impressed!

The template presentation with instructions is very neat also.

obilodeau:
Feature implemented in #93. Closing.

PR #93, opened 2016-09-08
Turning asciidoctor-revealjs into a proper asciidoctor back-end

obilodeau:
See #78. This will get updated as I progress.

### DIFF
--- a/asciidoctor-revealjs.gemspec
+++ b/asciidoctor-revealjs.gemspec
@@ -1,0 +1,31 @@
+# -*- encoding: utf-8 -*-
+require File.expand_path '../lib/asciidoctor-revealjs/version', __FILE__
+
+Gem::Specification.new do |s|
+  s.name = 'asciidoctor-revealjs'
+  s.version = Asciidoctor::Revealjs::VERSION
+  s.authors = ['Olivier Bilodeau']
+  s.email = ['olivier@bottomlesspit.org']
+  s.homepage = 'https://github.com/asciidoctor/asciidoctor-reveal.js'
+  s.summary = 'Converts AsciiDoc to HTML for a Reveal.js presentation'
+  s.description = 'An Asciidoctor converter that generates the HTML component of a Reveal.js presentation from AsciiDoc.'
+  s.license = 'MIT'
+  s.required_ruby_version = '>= 1.9.3'
+
+  files = begin
+    IO.popen('git ls-files -z') {|io| io.read }.split "\0"
+  rescue
+    Dir['**/*']
+  end
+  s.files = files.grep(/^(?:(?:bin|lib|templates)\/.+|Rakefile|(LICENSE|README)\.adoc)$/)
+
+  s.executables = ['asciidoctor-revealjs']
+  s.extra_rdoc_files = Dir['README.adoc', 'LICENSE.adoc', 'HACKING.adoc']
+  s.require_paths = ['lib']
+
+  s.add_runtime_dependency 'asciidoctor', '>= 1.5.4'
+  s.add_runtime_dependency 'slim', '~> 3.0.6'
+  s.add_runtime_dependency 'thread_safe', '~> 0.3.5'
+
+  s.add_development_dependency 'rake', '~> 10.4.2'
+end
--- a/asciidoctor-revealjs.gemspec
+++ b/asciidoctor-revealjs.gemspec
@@ -13,11 +13,12 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 1.9.3'
 
   files = begin
-    IO.popen('git ls-files -z') {|io| io.read }.split "\0"
+    (result = Open3.popen3('git ls-files -z') {|_, out| out.read }.split %(\0)).empty? ? Dir['**/*'] : result
   rescue
     Dir['**/*']
   end
-  s.files = files.grep(/^(?:(?:bin|lib|templates)\/.+|Rakefile|(LICENSE|README)\.adoc)$/)
+  s.files = files.grep %r/^(?:(?:lib|templates)\/.+|Gemfile|Rakefile|(?:CHANGELOG|LICENSE|README)\.adoc|#{s.name}\.gemspec)$/
+
 
   s.executables = ['asciidoctor-revealjs']
   s.extra_rdoc_files = Dir['README.adoc', 'LICENSE.adoc', 'HACKING.adoc']
--- a/lib/asciidoctor-revealjs/converter.rb
+++ b/lib/asciidoctor-revealjs/converter.rb
@@ -1,0 +1,30 @@
+require 'asciidoctor/converter/html5'
+require 'asciidoctor/converter/composite'
+require 'asciidoctor/converter/template'
+
+module Asciidoctor
+module Revealjs
+class Converter < ::Asciidoctor::Converter::CompositeConverter
+  ProvidedTemplatesDir = ::File.expand_path '../../../templates', __FILE__
+  register_for 'revealjs'
+
+  def initialize backend, opts = {}
+      # merge user templates with provided templates (user wins)
+      template_dirs = [ProvidedTemplatesDir]
+      if (user_template_dirs = opts[:template_dirs])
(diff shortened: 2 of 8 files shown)


obilodeau:
@mojavelinux I'll need some help here. So far I can bundle install the Gem, however when I try to run `asciidoctor-revealjs` I get the following error:

```
$ bundle exec asciidoctor-revealjs index.adoc --trace
bundler: failed to load command: asciidoctor-revealjs (/home/olivier/src/asciidoc/presentation-revealjs-starter/rubygems/ruby/2.3.0/bin/asciidoctor-revealjs)
NoMethodError: undefined method `gsub' for nil:NilClass
  /home/olivier/src/asciidoc/presentation-revealjs-starter/rubygems/ruby/2.3.0/bundler/gems/asciidoctor-reveal.js-853691e53158/templates/slim/section.html.slim:3:in `block in singleton class'
[...]
```

This is where the bug happens: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/slim/section.html.slim#L3

I think @id is nil. I tried removing the @ (which I've seen in bespoke) and it didn't work. Somehow the converter doesn't seem to be setting up the slim engine in the same way that using asciidoctor with -T would do.

obilodeau:
Created this https://github.com/obilodeau/presentation-revealjs-starter to test my Gem.

obilodeau:
While offline, I retried the package and I think it works after all... Will destroy, recreate and retest to validate.

asciidoctor-revealjs.gemspec
```diff
@@ -0,0 +1,31 @@
...
+  s.description = 'An Asciidoctor converter that generates the HTML component of a Reveal.js presentation from AsciiDoc.'
+  s.license = 'MIT'
+  s.required_ruby_version = '>= 1.9.3'
+
+  files = begin
+    IO.popen('git ls-files -z') {|io| io.read }.split "\0"
+  rescue
+    Dir['**/*']
+  end
+  s.files = files.grep(/^(?:(?:bin|lib|templates)\/.+|Rakefile|(LICENSE|README)\.adoc)$/)
```
mojavelinux: I've updated the logic for selecting files for the gem that is compatible across all Ruby platforms.

See https://github.com/asciidoctor/asciidoctor-bespoke/blob/6c24942259891819745f3a2f9dcc4828e6ba1fab/asciidoctor-bespoke.gemspec#L15-L20

mojavelinux:
> I think @id is nil. I tried removing the @ (which I've seen in bespoke) and it didn't work.

In general, we should stop using the low-level instance variable (with the @ prefix) and invoke the public accessor (without the @ prefix).

mojavelinux:
I've written up bullet-proof Bundler setup instructions:

## Prerequisites

If you're using RVM, make sure you switch away from any gemset:

    $ rvm use default

or

    $ rvm use system

(For some reason, when you use the system Ruby on Fedora, you also have to add the json gem to the Gemfile).

## Install

    $ rm -f Gemfile.lock
      bundle config --local github.https true
      bundle --path=.bundle/gems --binstubs=.bundle/.bin

## Run

    $ ./.bundle/.bin/asciidoctor-revealjs \
      -a revealjsdir=https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.3.0 index.adoc

or

    $ bundle exec asciidoctor-revealjs \
      -a revealjsdir=https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.3.0 index.adoc

## About

The approach here is to install gems into the project. This makes the setup far easier to understand and manage. (Similar to how node modules are managed). If something goes wrong, just delete the .bundle folder and start over.

I don't recommend putting .bundle/config into git because users may want to use a different setup. That's why I think it's appropriate just to put the instructions in the README.

mojavelinux:
I sent a PR to the starter project that includes these instructions.

obilodeau:
The starter presentation works now on my setup.

obilodeau:
Before we merge this, do you think these instructions should also be included in this repo? Or should we point the documentation to the starter repo and move it under the asciidoctor umbrella?

mojavelinux:
The starter project is probably where the instructions will be most helpful, but it doesn't hurt to mention it in both places. Perhaps what goes in this README is slightly reduced with a pointer to the starter repo for more details. But we'll know how much detail is needed as feedback comes in.

obilodeau:
Fixed conflict (with a rebase) and added instructions. I think this should be good to go.

obilodeau:
Made more documentation changes now that #95 was merged. Promoting both ruby and javascript back-ends in the top section before the ToC. I think both the node and bundler work improves our accessibility to new users a lot!

Waiting for a review before the merge.

lib/asciidoctor-revealjs/converter.rb
```diff
@@ -0,0 +1,30 @@
...
+  def initialize backend, opts = {}
+      # merge user templates with provided templates (user wins)
+      template_dirs = [ProvidedTemplatesDir]
+      if (user_template_dirs = opts[:template_dirs])
+        template_dirs += user_template_dirs.map {|d| ::File.expand_path d }
+      end
+      # create the main converter
+      template_converter = ::Asciidoctor::Converter::TemplateConverter.new backend,
+          template_dirs,
+          (opts.merge htmlsyntax: 'html', template_engine: 'slim')
```
Mogztter: If we want to use the same converter in Asciidoctor.js (cross-compiled with Opal), I think we need to provide an easy way to override this value from `slim` to `jade`

lib/asciidoctor-revealjs/converter.rb
```diff
@@ -0,0 +1,30 @@
+require 'asciidoctor/converter/html5'
+require 'asciidoctor/converter/composite'
+require 'asciidoctor/converter/template'
+
+module Asciidoctor
+module Revealjs
```
Mogztter: I'm not a Ruby code convention expert but I think this should be indented no ?

mojavelinux: I've recently discovered a trick that allows you to get the scoping behavior, but write this on a single line:

```ruby
module Asciidoctor; module Revealjs
  ...
end; end
```

That differs from the following where the scope of the intermediate modules are not visible:

```
module Asciidoctor::Revealjs
  ...
end
```

mojavelinux: (If you can tell, I don't like excessive indentation ;))

lib/asciidoctor-revealjs/converter.rb
mojavelinux: Indeed. We probably just want to detect the Opal environment here and set the correct value. I still consider the value to be a constant because this is about pointing to the source code, but we need to accommodate both environments.

obilodeau: While I agree, this is not obvious to me how it can be done. I looked at Opal's documentation for 5 minutes and didn't find anything. I don't think it should prevent the merge and you'll be able to propose a solution in a PR after the merge.

lib/asciidoctor-revealjs/converter.rb
obilodeau: Fixed

obilodeau:
May I proceed with the merge or you need more time?

lib/asciidoctor-revealjs/converter.rb
mojavelinux: We do this a ton in core. It's as simple as:

```ruby
if RUBY_ENGINE == 'opal'
  template_engine = 'jade'
else
  template_engine = 'slim'
end

...

(opts.merge htmlsyntax: 'html', template_engine: template_engine)
```

obilodeau: Done: ab3c719

lib/asciidoctor-revealjs/converter.rb
mojavelinux: Aside from the indentation issue, :+1:

mojavelinux:
It's up to you, @Mogztter. I want you to use what you feel is easiest.

Issue #88, opened 2016-09-05
Error generated if the asciidoctor source-highlighter: 'highlight.js' is not defined


cmoulliard:
This error is reported when we process the file using asciidoctor.js if the asciidoctor source-highlighter: 'highlight.js' attribute is not defined

![screenshot 2016-09-05 12 31 43](https://cloud.githubusercontent.com/assets/463790/18245966/9936f47c-7368-11e6-85e6-d455fb6186d9.png)

Maybe this code should be improved to define a default value 

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/jade/document.jade#L90

Mogztter:
I'm experiencing another issue. In my case `dependencies` attribute is:

```javascript
dependencies: [
            {
              src: 'reveal.js/lib/js/classList.js', condition: function () {
              return !document.body.classList;
            }
            },
            {},
            {
              src: 'reveal.js/plugin/zoom-js/zoom.js', async: true, condition: function () {
              return !!document.body.classList;
            }
            },
            {
              src: 'reveal.js/plugin/notes/notes.js', async: true, condition: function () {
              return !!document.body.classList;
            }
            }
          ]
```

The problem is that reveal.js does not work with an empty object:

```
reveal.js Uncaught TypeError: Cannot read property 'match' of undefined
```
https://github.com/hakimel/reveal.js/blob/539e774d31f91676bcc3f75e28168921cd27d819/js/reveal.js#L370

`s.src` is undefined

cmoulliard:
Maybe we should add nil like the slim template in this case --> https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/slim/document.html.slim#L173

cmoulliard:
This workaround which is not great works for jade --> Replace `{},` with `{ src: ''},`

mojavelinux:
The correct value is empty string. In the Slim/Ruby template, nil is equivalent to empty string. We definitely don't want to emit an empty JSON object. That's just asking for trouble.

cmoulliard:
You are right @mojavelinux, I have changed the line as such

```
#{(node.$attr('source-highlighter') == 'highlight.js') ? "{ src: '#{revealjsdir}/plugin/highlight/highlight.js', async: true, callback: function() { hljs.initHighlightingOnLoad(); } }," : ""}
```

I will submit a PR to include this modification

cmoulliard:
[...]

PR #95, opened 2016-09-14
Adds package.json

ggrossetie:
The goal is to make this backend easily available on the Node ecosystem.
I've created a simple converter for Jade templates (compiled in JavaScript with Opal) that will be available soon on `npmjs`: https://github.com/Mogztter/asciidoctor-template.js

When both projets will be published on `npmjs`, a user will be able to convert an AsciiDoc document into a Reveal.js presentation as follows:

``` sh
npm install asciidoctor.js # AFAIK this is only required on npm 2.x
npm install asciidoctor-reveal.js
```

``` javascript
// Loads Asciidoctor.js + Asciidoctor-template.js
var Asciidoctor = asciidoctor.Asciidoctor();
var Opal = asciidoctor.Opal;
Opal.load('nodejs');
Opal.load('pathname');
require('asciidoctor-template.js');

var options = Opal.hash({'safe': 'safe',
                         'template_dir': 'node_modules/asciidoctor-reveal.js/templates',
                         'backend': 'revealjs'});

var content = '...'; // AsciiDoc content
var result = Asciidoctor.$convert(content, options); // Congrats you have a Reveal.js presentation !
```

Feel free to add authors and a better description :smile:

### DIFF
--- a/HACKING.adoc
+++ b/HACKING.adoc
@@ -1,6 +1,7 @@
 = HACKING asciidoctor-reveal.js
 :toc: preamble
 :toclevels: 2
+:uri-nodejs-download: https://nodejs.org/en/download/
 
 Short instructions that aim to help potential contributors.
 
@@ -190,3 +191,94 @@ You might even be able to get some feedback early which could save you some time
 You can apply that label to a pull request that is complete and ready for review.
 
 Makes triaging easier.
+
+
+== Node package
+
+=== Initial setup
+
+First you must install and configure {uri-nodejs-download}[Node.js] on your machine.
+
+=== Setup a test project
+
+In order to test the Node package, you need to create a test project adjacent to the clone of the `asciidoctor-reveal.js` repository:
+
+ $ mkdir test-project
+ $ cd test-project
+ $ npm init -y
+
+Now, install the dependencies:
+
+ $ npm i --save asciidoctor.js@1.5.5-2
+ $ npm i --save asciidoctor-template.js@1.5.5-2
+ $ npm i --save ../asciidoctor-reveal.js
+
+IMPORTANT: As mentioned above, the `asciidoctor-reveal.js` repository must be cloned next to the `test-projet` directory.
+
+Create an AsciiDoc file:
+
+[source]
+----
+$ cat <<- EOF > presentation.adoc
+= Title Slide
+// depending on your npm version, you might have to try the other 'revealjsdir' value.
+//:revealjsdir: node_modules/reveal.js
+
+== Slide One
+
+* Foo
+* Bar
+* World
+
+== Slide Two
+
+Hello World - Good Bye Cruel World
+
+[NOTE.speaker]
+--
+Actually things aren't that bad
+--
+EOF
+----
+
+Use the Asciidoctor API to convert an AsciiDoc file into a Reveal.js presentation:
+
+[source]
+----
+$ cat <<- 'EOF' > asciidoctor-revealjs.js
+// Load asciidoctor.js + asciidoctor-template.js
+var asciidoctor = require('asciidoctor.js')();
+var Asciidoctor = asciidoctor.Asciidoctor();
+var Opal = asciidoctor.Opal;
+Opal.load('nodejs');
+Opal.load('pathname');
+require('asciidoctor-template.js');
+
+// Convert the document 'presentation.adoc' using Reveal.js backend
+var attributes = 'revealjsdir=node_modules/asciidoctor-reveal.js/node_modules/reveal.js';
+var options = Opal.hash({safe: 'safe',
+                         backend: 'revealjs',
+                         attributes: attributes});
+Asciidoctor.$convert_file('presentation.adoc', options);
+EOF
+----
+
+Run the following command:
+
+ $ node asciidoctor-revealjs.js
+
+A file named `presentation.html` has been created in the `test-project` directory.
+You can open this file in your browser and enjoy!
+
+=== Publish a new version
+
+. Update the "version" field in `package.json`
+. Commit and push your changes
+. Create a tag
+. Using npm:
++
+ $ npm login
+ $ npm publish
+
+. Check that the new version is available on https://www.npmjs.com/package/asciidoctor-reveal.js[npmjs.com]
+
--- a/README.adoc
+++ b/README.adoc
@@ -1,6 +1,7 @@
 = reveal.js backend for Asciidoctor
 :toc: preamble
 :toclevels: 2
+:uri-nodejs-download: https://nodejs.org/en/download/
 
 The https://github.com/asciidoctor/asciidoctor-reveal.js[reveal.js backend] is a collection of Slim templates for https://github.com/asciidoctor/asciidoctor[Asciidoctor] that transform an AsciiDoc document into HTML5 slides animated by http://lab.hakim.se/reveal-js/[reveal.js].
 
@@ -8,6 +9,8 @@ The https://github.com/asciidoctor/asciidoctor-reveal.js[reveal.js backend] is a
 
 
 == Prerequisites
+
+NOTE: Since Asciidoctor can run in JavaScript, we also provide a <<node-package>>.
 
 . Ensure Asciidoctor, Slim and their dependencies are installed:
 
@@ -20,7 +23,6 @@ The https://github.com/asciidoctor/asciidoctor-reveal.js[reveal.js backend] is a
 . Copy or clone presentation library (to output destination/branch)
 
   $ git clone -b 3.0.0 git://github.com/hakimel/reveal.js.git
-
 
 == Rendering the AsciiDoc into slides
 
@@ -35,6 +37,39 @@ NOTE: If you want to use reveal.js 2, please switch your asciidoctor-reveal.js r
 
 TIP: If you are using https://pages.github.com/[GitHub Pages], plan ahead by keeping your source files on `master` branch and all output files on the `gh-pages` branch.
 
+[[node-package]]
+== Node package
+
+First you must install and configure {uri-nodejs-download}[Node.js] on your machine.
+
+Using npm:
+
+ $ npm i --save asciidoctor.js@1.5.5-2
+ $ npm i --save asciidoctor-template.js@1.5.5-2
+ $ npm i --save asciidoctor-reveal.js@1.5.5-2
+
+Once the package is installed, you can convert AsciiDoc files using Reveal.js backend.
+Here we are converting a file named `presentation.adoc` into a Reveal.js presentation using Node.js:
+
+.asciidoctor-revealjs.js
+[source, javascript]
+----
+// Load asciidoctor.js + asciidoctor-template.js
+var asciidoctor = require('asciidoctor.js')();
+var Asciidoctor = asciidoctor.Asciidoctor();
+var Opal = asciidoctor.Opal;
+Opal.load('nodejs');
+Opal.load('pathname');
+require('asciidoctor-template.js');
+
+// Convert the document 'presentation.adoc' using Reveal.js backend
+var attributes = 'revealjsdir=node_modules/asciidoctor-reveal.js/node_modules/reveal.js';
+var options = Opal.hash({safe: 'safe',
+                         backend: 'revealjs',
+                         attributes: attributes});
+Asciidoctor.$convert_file('presentation.adoc', options); // <1>
+----
+<1> Create a file named `presentation.html` (in the directory where command is run)
 
 == Some Examples
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 test/*.html
 test/reveal.js/
 node_modules/
+npm-debug.log
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 test/*.html
 test/reveal.js/
+node_modules/
--- a/package.json
+++ b/package.json
@@ -1,0 +1,40 @@
+{
+  "name": "asciidoctor-reveal.js",
+  "version": "1.5.5-2",
+  "description": "Reveal.js backend templates for Asciidoctor.js, implemented in Jade",
+  "engines": {
+    "node": ">=0.12"
+  },
+  "files": [
+    "templates/jade",
+    "LICENSE.adoc",
+    "README.adoc"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/asciidoctor/asciidoctor-reveal.js.git"
+  },
+  "keywords": [
+    "asciidoc",
+    "asciidoctor",
+    "opal",
+    "javascript",
+    "library",
+    "backend",
+    "template",
+    "reveal.js",
+    "jade"
+  ],
+  "authors": [
+    "Guillaume Grossetie (https://github.com/mogztter)"
+  ],
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/asciidoctor/asciidoctor-reveal.js/issues"
+  },
+  "homepage": "https://github.com/asciidoctor/asciidoctor-reveal.js",
+  "dependencies": {
+    "reveal.js": "3.3.0",
+    "asciidoctor-template.js": "1.5.5-2"
+  }
+}
--- a/templates/jade/document.jade
+++ b/templates/jade/document.jade
@@ -2,6 +2,7 @@ doctype html
 html(lang=node.document.$attr("lang", "en") )
   head
     meta(charset="UTF-8")
+    - revealjsdir = node.$attr('revealjsdir', 'reveal.js')
     each key in ["description","keywords","author","copyright"]
       if node.$attr(key, false)
         meta(name=key,content=node.$attr(key))
@@ -10,21 +11,21 @@ html(lang=node.document.$attr("lang", "en") )
     meta(content="black-translucent",name="apple-mobile-web-app-status-bar-style")
     meta(content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, minimal-ui",name="viewport")
     link(href="//maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css",rel="stylesheet")
-    link(href="reveal.js/css/reveal.css",rel="stylesheet")
+    link(href="#{revealjsdir}/css/reveal.css",rel="stylesheet")
     if node.$attr("revealjs_customtheme", false)
       link(rel='stylesheet',href="#{node.$attr('revealjs_customtheme')}",id='theme')
     else
-      link(rel='stylesheet',href='reveal.js/css/theme/serif.css',id='theme')
-    link(href="reveal.js/lib/css/zenburn.css",rel="stylesheet")
+      link(rel='stylesheet',href='#{revealjsdir}/css/theme/serif.css',id='theme')
+    link(href="#{revealjsdir}/lib/css/zenburn.css",rel="stylesheet")
     script(src = "http://getfirebug.com/firebug-lite.js#startOpened=false")
     script(type='text/javascript').
-      document.write('<link rel="stylesheet" href="reveal.js/css/print/' + ( window.location.search.match(/print-pdf/gi) ? 'pdf' : 'paper' ) + '.css" type="text/css" media="print">');
+      document.write('<link rel="stylesheet" href="#{revealjsdir}/css/print/' + ( window.location.search.match(/print-pdf/gi) ? 'pdf' : 'paper' ) + '.css" type="text/css" media="print">');
   body
     .reveal
       .slides
         !{node.$content()}
-    script(src = "reveal.js/lib/js/head.min.js")
-    script(src = "reveal.js/js/reveal.js")
+    script(src = "#{revealjsdir}/lib/js/head.min.js")
+    script(src = "#{revealjsdir}/js/reveal.js")
     script(type='text/javascript').
       function initializeReveal() {
         // See https://github.com/hakimel/reveal.js#configuration for a full list of configuration options
@@ -83,20 +84,19 @@ html(lang=node.document.$attr("lang", "en") )
           // Optional libraries used to extend on reveal.js
           dependencies: [
             {
-              src: 'reveal.js/lib/js/classList.js', condition: function () {
-              return !document.body.classList;
-            }
+              src: '#{revealjsdir}/lib/js/classList.js',
+              condition: function () { return !document.body.classList; }
             },
-            #{(node.$attr('source-highlighter') == 'highlight.js') ? "{ src: 'reveal.js/plugin/highlight/highlight.js', async: true, callback: function() { hljs.initHighlightingOnLoad(); } }," : ""}
+            #{(node.$attr('source-highlighter') == 'highlight.js') ? "{ src: '#{revealjsdir}/plugin/highlight/highlight.js', async: true, callback: function() { hljs.initHighlightingOnLoad(); } }," : ""}
             {
-              src: 'reveal.js/plugin/zoom-js/zoom.js', async: true, condition: function () {
-              return !!document.body.classList;
-            }
+              src: '#{revealjsdir}/plugin/zoom-js/zoom.js',
+              async: true,
+              condition: function () { return !!document.body.classList; }
             },
             {
-              src: 'reveal.js/plugin/notes/notes.js', async: true, condition: function () {
-              return !!document.body.classList;
-            }
+              src: '#{revealjsdir}/plugin/notes/notes.js',
+              async: true,
+              condition: function () { return !!document.body.classList; }
             }
           ]
         });


mojavelinux:
This is good. In the future, though, I'd really like to set a goal that the converter can be loaded through the backend option instead of referencing the templates directly. That's a matter of creating a Ruby or JavaScript-based converter class that sets up the template (and/or composite) converter and points it at the templates.

Mogztter:
Since I'm using a [custom template converter](https://github.com/Mogztter/asciidoctor-template.js/blob/master/lib/asciidoctor/core_ext/template.rb) I can add a default value when running on Node:

```ruby
if backend == 'revealjs' && running_node
  template_dir = template_dir || 'node_modules/asciidoctor-reveal.js/templates'
end
```

@mojavelinux Is this what you have in mind ?

mojavelinux:
That's a good interim step. I'd still like to see a custom converter introduced to manage the wiring because it can do other things such as set the basebackend and htmlsyntax attributes. In other words, official converters should always be registered formally with the API. Using templates "ad-hoc" is only meant to be an end-user customization.

Mogztter:
I've just noticed that @obilodeau is working on an "official" converter registered with the API.
https://github.com/asciidoctor/asciidoctor-reveal.js/pull/93

But AFAIK this won't work in a JavaScript environment.

Currently there's multiple issues with template converter backend in JavaScript:
- it's hard to locate and load files on various JavaScript environments (especially when running in a browser)
- composite converter is not working (didn't dig too much)
- requires complex setup (git clone, require(), template_dir...)

But I still think that the template converter backend is great for end user customization. 

However "official" converters should be easier to use/integrate. Maybe we should rewrite this converter in Ruby like the html5 converter ? But that's another subject :)

cmoulliard:
> Is this what you have in mind ?

I like the by default value that you suggest

obilodeau:
For me this is a complex change to handle since I don't fully understand all of the stuff that is going on. That said, I'm willing to integrate anything that makes using AsciiDoc (including asciidoc[tor] + reveal.js) simpler as long as someone is around when issues arises.

However there are two things I want to say here:

First, you need to add some documentation about it in the README. Even if I would like to test what you propose I have no idea how to do so. Some doc would help me (and others) for sure.

Second is that I'll let @mojavelinux give the final :+1: on this one. Since he's the one who understands all the moving pieces the most.

Since reveal.js is a node project I think this can't be a bad thing to have this back-end published over there as well. Thanks for that!

Mogztter:
> For me this is a complex change to handle since I don't fully understand all of the stuff that is going on. That said, I'm willing to integrate anything that makes using AsciiDoc (including asciidoc[tor] + reveal.js) simpler as long as someone is around when issues arises.

Sure, that's fair enough, I will be around :wink: 

> First, you need to add some documentation about it in the README. Even if I would like to test what you propose I have no idea how to do so. Some doc would help me (and others) for sure.

I didn't know how to integrate with the current README, I think we will need to restructure the sections a bit. Currently the "Prerequisites" and "Rendering the AsciiDoc into slides" sections explain how to use Reveal.js templates with Asciidoctor Ruby.
So we will need to create a section/subsection to explain how to setup and use this backend with Asciidoctor.js. Wdyt ?

> Second is that I'll let @mojavelinux give the final :+1: on this one. Since he's the one who understands all the moving pieces the most.

Ok

> Since reveal.js is a node project I think this can't be a bad thing to have this back-end published over there as well. Thanks for that!

:+1:

obilodeau:
> I didn't know how to integrate with the current README, I think we will need to restructure the sections a bit. Currently the "Prerequisites" and "Rendering the AsciiDoc into slides" sections explain how to use Reveal.js templates with Asciidoctor Ruby.
> So we will need to create a section/subsection to explain how to setup and use this backend with Asciidoctor.js. Wdyt ?

Yes, do some intro like: there are two ways to use this back-end with varying levels of functionality. Then put all the ruby specific in a section and the javascript specific in another. Don't worry too much about phrasing, we can fix that later. As long as the steps are there for me to try the back-end.

cmoulliard:
I think that this is a good idea to setup this `revealjs_base_dir` as most probably revealjs will be installed separately from the templates and should be available from the HTML folder containing the Revealjs slideshow.

obilodeau:
Can't you just tell the user how to configure the existing `:revealjsdir:`
attribute in the documentation?

Most people I know who use revealjs fork it to add a custom theme. Using
the attribute is the way to do it.

obilodeau:
It's not up to the template engine. AsciiDoc document attributes are available in both engines as it is provided by the core. It's just that the Jade template designer chose not to use it (for a reason he didn't explain).

Task added to #63. 

`revealjs_base_dir` vs `revealjsdir`

Please absolutely refrain from introducing new revealjs specific attributes if they already exist for slim. This will only make multi-engine maintenance more complicated.

Mogztter:
I didn't know about `revealjsdir`, I'm fine with the naming.

> It's not up to the template engine. AsciiDoc document attributes are available in both engines as it is provided by the core.

When I said "only defined", I meant "only used" in Slim templates.

>  It's just that the Jade template designer chose not to use it (for a reason he didn't explain).

I think it's not on purpose since I don't see any real drawback... ?

obilodeau:
Agreed. Not knowing it existed is a perfectly valid reason not to use it. I
just assumed (wished) he took inspiration from the slim template.

README.adoc
```diff
@@ -35,6 +36,33 @@ NOTE: If you want to use reveal.js 2, please switch your asciidoctor-reveal.js r
 
 TIP: If you are using https://pages.github.com/[GitHub Pages], plan ahead by keeping your source files on `master` branch and all output files on the `gh-pages` branch.
 
+[[node-package]]
+== Node package
+
+Using npm:
+
+ $ npm i -g npm
```
obilodeau: Disclaimer: I'm a node/javascript n00b.

Why the global install of npm (with npm!?)? On my arch linux setup npm is managed by my package manager and I don't want to change that.

README.adoc
```diff
@@ -35,6 +36,33 @@ NOTE: If you want to use reveal.js 2, please switch your asciidoctor-reveal.js r
...
+ $ npm i --save asciidoctor.js
+ $ npm i --save asciidoctor-reveal.js
```
obilodeau:
```
npm ERR! 404 Registry returned 404 for GET on https://registry.npmjs.org/asciidoctor-reveal.js
```

So how can I test locally?

I tried using `npm link`, even as root (which I hate doing) and it always failed because of a missing dependency on `asciidoctor-template.js`:

```
npm ERR! 404  'asciidoctor-template.js' is not in the npm registry.
```

So I tried doing a `sudo npm link` inside a checkout of asciidoctor-template.js and it worked but a subsequent `sudo npm link` from asciidoctor-reveal.js still failed like if asciidoctor-template.js was not installed:

```
npm ERR! 404  'asciidoctor-template.js' is not in the npm registry.
```

Sorry for my lack of experience but I tried. Can you help me here?

README.adoc
```diff
@@ -35,6 +36,33 @@ NOTE: If you want to use reveal.js 2, please switch your asciidoctor-reveal.js r
...
+
+In Node.js:
```
obilodeau: You provide the filename and the content, which is good but then how do you invoke the actual rendering? It is unclear to me.

README.adoc
```diff
@@ -35,6 +36,33 @@ NOTE: If you want to use reveal.js 2, please switch your asciidoctor-reveal.js r
...
+var Opal = asciidoctor.Opal;
+Opal.load('nodejs');
+Opal.load('pathname');
+require('asciidoctor-template.js');
+
+// Convert the document 'presentation.adoc' using Reveal.js backend
+var options = Opal.hash({safe: 'safe',
+                         template_dir: 'node_modules/asciidoctor-reveal.js/templates',
+                         backend: 'revealjs'});
+var result = Asciidoctor.$convert_file('presentation.adoc', options);
```
obilodeau: The fact that `presentation.adoc` is a filename and that it is related to what you are trying to render is important and should be clarified in the documentation outside the code listing.

obilodeau (changes requested):
I am not able to test since the packages are not in the npm registry. See inline comments. I suggest you provide instructions on how to test the whole thing in `HACKING.adoc`.

For me what is important as an outsider to the javascript/node ecosystem is what I can independently reproduce with your instructions (and with the least effort).

Thanks for taking the time to educate me and others! Improvements to the documentation will increase the number of users and ease the path to contributors.

cmoulliard:
Here are the bash instructions that you can use to create and test a node project using Guillaume's modifications. 

Instructions : Create a bash file, change the permissions (chmod +x file) & execute it within a folder

```
#!/usr/bin/env bash

echo "## Cleanup & create folder"
rm -rf lib/
rm -rf reveal.js-3.3.0
rm -rf node_modules/
rm reveal.js.zip
rm *.js
rm *.json
rm *.adoc
rm *.html
rm *.log

mkdir lib
cd lib 

echo "## Clone Asciidoctor-template, install node modules"
git clone https://github.com/Mogztter/asciidoctor-template.js.git
cd asciidoctor-template.js
npm install
npm run dist
cd ..

echo "## Clone Asciidoctor-revealjs"
git clone https://github.com/Mogztter/asciidoctor-reveal.js.git
cd asciidoctor-reveal.js
git checkout node-package
npm install ../asciidoctor-template.js
npm install
cd ../..

echo "## Create files adoc, package"
cat <<- EOF > presentation.adoc
= Title Slide

== Slide One

* Foo
* Bar
* World

== Slide Two

Hello World - Good Bye Cruel World

[NOTE.speaker]
--
Actually things aren't that bad
--
EOF

cat <<- EOF > package.json
{
  "name": "ascii-revealjs",
  "version": "1.0.0",
  "description": "Asciidoctro Revealjs demo",
  "main": "asciidoctor-reveal.js",
  "scripts": {
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "asciidoctor.js": "^1.5.5-2"
  }
}
EOF

cat <<- 'EOF' > reveal-ascii.js
// Load asciidoctor.js + asciidoctor-template.js
var Asciidoctor = asciidoctor.Asciidoctor();
var Opal = asciidoctor.Opal;
Opal.load('nodejs');
Opal.load('pathname');
require('asciidoctor-template.js');

// Convert the document 'presentation.adoc' using Reveal.js backend
var options = Opal.hash({safe: 'safe',
                         template_dir: 'node_modules/asciidoctor-reveal.js/templates',
                         backend: 'revealjs'});
var result = Asciidoctor.$convert_file('presentation.adoc', options);
EOF

echo "## Install node modules"
npm install --safe ./lib/asciidoctor-template.js/
npm install --safe ./lib/asciidoctor-reveal.js/

echo "## Download revealjs"
curl -L https://github.com/hakimel/reveal.js/archive/3.3.0.zip > reveal.js.zip
unzip reveal.js.zip
ln -s reveal.js-3.3.0 reveal.js

echo "## Generate Slideshow"
node reveal-ascii.js
```

README.adoc
Mogztter: Sorry about that, I will publish a "beta" version in the npm registry, to make things easier.

README.adoc
Mogztter: I took inspiration from https://github.com/lodash/lodash#installation, the only hard requirement is to have npm available on your PATH. So basically if you install Node.js you are good to go.

README.adoc
Mogztter: Ok

Mogztter:
Wow thanks @cmoulliard :smile: 

@obilodeau I will try to make the requested changes ASAP

obilodeau:
> Here are the bash instructions that you can use to create and test a node project using Guillaume's modifications.

Thanks! For reference, I needed to do another `npm install` after the two `install --safe` otherwise it crashed because I'm missing asciidoctor-js:

```
$ node reveal-ascii.js
module.js:457
    throw err;
    ^

Error: Cannot find module 'asciidoctor.js'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/olivier/src/asciidoc/presentation-asciidoctorjs-test/reveal-ascii.js:2:44)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
```

After this the render worked:

```
$ npm install
ascii-revealjs@1.0.0 /home/olivier/src/asciidoc/presentation-asciidoctorjs-test
└── asciidoctor.js@1.5.5-2 

npm WARN ascii-revealjs@1.0.0 No repository field.

$ node reveal-ascii.js
Warning: missing space before text for line 25 of jade file "undefined"
Warning: missing space before text for line 8 of jade file "undefined"
Warning: missing space before text for line 10 of jade file "undefined"
Warning: missing space before text for line 15 of jade file "undefined"
Warning: missing space before text for line 21 of jade file "undefined"
Warning: missing space before text for line 23 of jade file "undefined"
Warning: missing space before text for line 25 of jade file "undefined"
Warning: missing space before text for line 8 of jade file "undefined"
Warning: missing space before text for line 10 of jade file "undefined"
Warning: missing space before text for line 15 of jade file "undefined"
Warning: missing space before text for line 14 of jade file "undefined"
Warning: missing space before text for line 17 of jade file "undefined"
```

Note that I needed to work around the empty `{ }` bug otherwise reveal.js would crash.

The render is missing a title page and has other problems but those were expected and does not prevent us from merging this.

obilodeau:
The bash script should be turned into instructions and added into `HACKING.adoc` for documentation purposes. Also, I would like to understand what is the purpose of `asciidoctor-template.js` in a one-liner.

Mogztter:
> Also, I would like to understand what is the purpose of asciidoctor-template.js in a one-liner.

What do you mean by "in a one-liner" ? Basically, `asciidoctor-template.js` is a JavaScript implementation of the ["official" template converter written in Ruby](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/template.rb).
I made some changes mostly to remove Slim/Tilt/Erb specific code and use Jade instead of Slim/Erb as a template engine.

The template converter role is:

 1. [scan the template directories to find the template file by name](https://github.com/Mogztter/asciidoctor-template.js/blob/ad241151f15b7301ba4f01c25cedbaedbd4007d6/lib/asciidoctor/core_ext/template.rb#L14)
 2. [convert Asciidoctor node using the template found (using Jade as a template engine)](https://github.com/Mogztter/asciidoctor-template.js/blob/ad241151f15b7301ba4f01c25cedbaedbd4007d6/lib/asciidoctor/core_ext/template.rb#L41)

If you are wondering, [lib/asciidoctor/core_ext/factory.rb](https://github.com/Mogztter/asciidoctor-template.js/blob/ad241151f15b7301ba4f01c25cedbaedbd4007d6/lib/asciidoctor/core_ext/factory.rb) is a monkey-patch of [lib/asciidoctor/converter/factory.rb](https://github.com/asciidoctor/asciidoctor/blob/60dfaa0e05b10ef5e1c807554270f2a10ceaf71a/lib/asciidoctor/converter/factory.rb#L223) to remove the `CompositeConverter` (because it's not working yet in a JavaScript environment)

Does that make sense ? :grinning: :thought_balloon:

README.adoc
obilodeau: I suggest you simply add a Prerequisite section with Node.js then. Link it to the website. This way people will figure out how to install it on their own based on their requirement, preference, system, knowledge, etc.

obilodeau:
Once the README is fixed up and test setup instructions are written in `HACKING.adoc` I'll merge this. Thanks!

obilodeau:
My `package.json` was modified during my tests. Here's the diff:

```diff
$ git diff
diff --git a/package.json b/package.json
index b109129..294cb0d 100644
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
   },
   "homepage": "https://github.com/asciidoctor/asciidoctor-reveal.js",
   "dependencies": {
-    "reveal.js": "3.3.0",
-    "asciidoctor-template.js": "1.5.5-2"
+    "asciidoctor-template.js": "1.5.5-2",
+    "asciidoctor.js": "^1.5.5-2",
+    "reveal.js": "3.3.0"
   }
 }
```

I don't know if this is expected or not?

Mogztter:
@obilodeau Done, let me know if this is OK

I've added a section on `HACKING.adoc` to explain how to setup a test project and how to publish a new version on npmjs.com

Mogztter:
> My package.json was modified during my tests. Here's the diff, I don't know if this is expected or not?

You are not supposed to run `npm i --save asciidoctor.js@1.5.5-2` command in the `asciidoctor-reveal.js` repository. If you use `--save` option, `npm` will save the dependency in the `package.json` that's why your `package.json` was modified.

> I've also fixed this issue #88 and used revealjsdir in document.jade.

If you want I can create separated pull requests for these two issues.

mojavelinux:
The location of reveal.js seems to depend on which version of Node you are using. If you use Node v4, then reveal.js ends up in *node_modules/asciidoctor-reveal.js/node_modules/reveal.js*. However, if you use v0.12, v5 or v6, then it ends up in *node_modules/reveal.js*. I think we should recommend against using v4 because it has different behavior.

Perhaps it would be a good idea to switch from Jade to Pug. As far as I understand, it's just a matter of renaming the files with the .pug extension and updating the dependency.

mojavelinux:
Aside from that, this is amazing!

mojavelinux:
We should encourage users to use Node v6 since that's the stable version. We can recommend using nvm.io to manage node versions. Honestly, if you aren't using nvm, I don't know how you are keeping your sanity anyway.

mojavelinux:
We should recommend using the `-y` flag on the `npm init` command. Otherwise, all those questions are just confusing for a first-time user.

HACKING.adoc
```diff
@@ -188,3 +189,92 @@ You might even be able to get some feedback early which could save you some time
...
+Hello World - Good Bye Cruel World
+
+[NOTE.speaker]
+--
+Actually things aren't that bad
+--
+EOF
+----
+
+Use the Asciidoctor API to convert an AsciiDoc file into a Reval.js presentation:
```
mojavelinux: Reveal.js is spelled incorrectly in this sentence.

HACKING.adoc
```diff
@@ -188,3 +189,92 @@ You might even be able to get some feedback early which could save you some time
...
+ $ cd test-project
+ $ npm init
+
+Now, install the dependencies:
+
+ $ npm i --save asciidoctor.js@1.5.5-2
+ $ npm i --save asciidoctor-template.js@1.5.5-2
+ $ npm i --save ../asciidoctor-reveal.js
+
+IMPORTANT: Here we assume that the repository `asciidoctor-reveal.js` is cloned next to the `test-projet`.
```
mojavelinux: The assumptions about the project layout should be stated before the commands.

HACKING.adoc
```diff
@@ -188,3 +189,92 @@ You might even be able to get some feedback early which could save you some time
...
+
+== Node package
+
+=== Initial setup
+
+First you must install and configure {uri-nodejs-download}[Node.js] on your machine.
+
+=== Setup a test project
+
+In order to test the Node package, you will need to create a test project:
```
mojavelinux: It should be clear where this folder should be created. It should be created adjacent to the clone of asciidoctor-reveal.js.

mojavelinux:
I noticed the default language on the HTML root is "tr" instead of "en". That should be switched back.

mojavelinux:
I think to keep the Pug templates inline with the Slim templates, we'll definitely want to look into getting asciidoctor-doctest setup. With this converter, that should now be possible. We may have to inform asciidoctor-doctest how to invoke it...but once the HTML is generated, the rest should be easy to follow.

mojavelinux:
I want to say again, this is pretty incredible. It also brings hope to running the Bespoke.js converter in JavaScript.

mojavelinux:
There will need to be two official Reveal.js converters, one for each platform, since it's necessary to extend from different converter classes. For the JavaScript platform, the converter will need to extend @Mogztter's template converter, then setup things like the basebackend and template_dirs and such.

The only think missing from the template converter for JavaScript is the fallback to the built-in HTML5 converter. It's not critical since we provide all the templates in this case, but to be consistent with the built-in template converter, we'll want to add that at some point.

@Mogztter we should definitely get that template converter into the Asciidoctor organization. I've opened up transfer rights. You can either bring in that repository or merge it into asciidoctor.js. It's up to you how you want to manage it.

Mogztter:
> The location of reveal.js seems to depend on which version of Node you are using. If you use Node v4, then reveal.js ends up in node_modules/asciidoctor-reveal.js/node_modules/reveal.js. However, if you use v0.12, v5 or v6, then it ends up in node_modules/reveal.js. I think we should recommend against using v4 because it has different behavior.

Are you sure about that ? I can't reproduce this behavior. When using both Node 4.2.6 and 0.12.0, `reveal.js` is available at `node_modules/asciidoctor-reveal.js/node_modules/reveal.js`.

> Perhaps it would be a good idea to switch from Jade to Pug. As far as I understand, it's just a matter of renaming the files with the .pug extension and updating the dependency.

I wanted to switch from Jade to Pug but Pug is still in beta and the latest stable version is [Jade@1.11.0](https://github.com/pugjs/pug/tree/1.11.0).

> Before the renaming, we had already begun working on an incompatible Jade 2.0.0. We have then made it so that this new major version bump will coincide with the rename to Pug. Therefore, upgrading from Jade to Pug will be the same process as upgrading any other package with a major version bump. Currently, Pug 2.0.0 is still under beta stage, and there are several syntactic differences we have deprecated and removed. Such differences are documented at #2305.

https://github.com/pugjs/pug/tree/2.0.0-beta6#rename-from-jade

HACKING.adoc
Mogztter: Oups, fixed :disappointed:

HACKING.adoc
Mogztter: I've added a statement before the commands (in addition to the admonition)

HACKING.adoc
Mogztter: Ok

Mogztter:
> I noticed the default language on the HTML root is "tr" instead of "en". That should be switched back.

Good catch, I've created a pull request #103

Mogztter:
> I want to say again, this is pretty incredible. It also brings hope to running the Bespoke.js converter in JavaScript.

Soon :+1:

Mogztter:
> There will need to be two official Reveal.js converters, one for each platform, since it's necessary to extend from different converter classes. For the JavaScript platform, the converter will need to extend @Mogztter's template converter, then setup things like the basebackend and template_dirs and such.

I agree. For now I've defined `template_dirs` for Reveal.js [here](https://github.com/Mogztter/asciidoctor-template.js/blob/master/lib/asciidoctor/core_ext/factory.rb#L40) but that's a short term solution until we have a proper Node build available in this repository.

> The only think missing from the template converter for JavaScript is the fallback to the built-in HTML5 converter. It's not critical since we provide all the templates in this case, but to be consistent with the built-in template converter, we'll want to add that at some point.

Yes, as mentioned earlier, `CompositeConverter` is not working in a JavaScript environment. I don't quite remember the root cause but I need to figure it out.

> @Mogztter we should definitely get that template converter into the Asciidoctor organization. I've opened up transfer rights. You can either bring in that repository or merge it into asciidoctor.js. It's up to you how you want to manage it.

Ok :+1:

cmoulliard (commented):
We must explain that when we install asciidoctor-reveal.js node module (= templates + test code), then the project reveal.js (css, js, ..) will be installed too as module and how we can access it using revealjsdir attr. As the end user could be also interested about the reveal.js version used, this info should also be documented somewhere (if not yet done).

obilodeau:
The HACKING instructions are good, the README changes are ok and I'm willing to merge before this PR has too many comments.

There's only one thing. On my machine, reveal.js was installed in `node_modules/reveal.js/` and not `node_modules/asciidoctor-reveal.js/node_modules/reveal.js/`. Removing attributes from the `asciidoctor-revealjs.js` file and adding this line to the `presentation.adoc` made it work:

```
:revealjsdir: node_modules/reveal.js/
```

Running node version v6.6.0 and npm v3.10.7 (if that's relevant).

What I suggests:

Change your `presentation.adoc` file to this:

```asciidoc
= Title Slide
// depending on your node.js version, you might have to try the other
// revealjsdir value
//:revealjsdir: node_modules/reveal.js/
:revealjsdir: node_modules/asciidoctor-reveal.js/node_modules/reveal.js/

== Slide One

* Foo
* Bar
* World

== Slide Two

Hello World - Good Bye Cruel World

[NOTE.speaker]
--
Actually things aren't that bad
--
```

This way there is an easily discoverable way to fix the problem and you educate the user about the `revealjsdir` attribute. This attribute is probably the most important one for anyone modifying the default reveal.js look and feel so I don't think it is unreasonable to introduce it at this point.

obilodeau (approved):

Mogztter:
Yes that's a good idea.
I think this is related to the npm version, not the Node.js version.
I will make this change today. Thanks Olivier.

Le 27 sept. 2016 6:03 AM, "Olivier Bilodeau" <notifications@github.com> a
écrit :

cmoulliard:
Can somebody resolves the conflicts and apply the PR please ?

Mogztter:
> Can somebody resolves the conflicts and apply the PR please ?

Done!

> I think this is related to the npm version, not the Node.js version.

Confirmed, this is related to the `npm` version.
I think we should reintroduce the commande `npm i -g npm` to update npm version. 

I'm currently running version `3.10.8` (with Node.js 0.12) and there's a ton of improvements. I may have to change the default value of `revealjsdir` because the latest version of `npm` uses a flat hierarchy and `reveal.js` is at the root of `node_modules` .

obilodeau:
> I think we should reintroduce the commande npm i -g npm to update npm version.

I tend to disagree. Running this as root overwrites files that should be under the authority of my package manager. This makes package integrity verification impossible which I don't accept with from a security standpoint.

README.adoc
```diff
@@ -35,6 +37,39 @@ NOTE: If you want to use reveal.js 2, please switch your asciidoctor-reveal.js r
...
+var Opal = asciidoctor.Opal;
+Opal.load('nodejs');
+Opal.load('pathname');
+require('asciidoctor-template.js');
+
+// Convert the document 'presentation.adoc' using Reveal.js backend
+var attributes = 'revealjsdir=node_modules/asciidoctor-reveal.js/node_modules/reveal.js';
+var options = Opal.hash({safe: 'safe',
+                         backend: 'revealjs',
+                         attributes: attributes});
```
obilodeau: Passing attributes as arguments makes overriding them in the document impossible. See [the documentation on attribute assignment precedence](http://asciidoctor.org/docs/user-manual/#attribute-assignment-precedence) for more information about that.

I suggest that you push the `revealjsdir` attribute default value in the example document as I suggested earlier.

HACKING.adoc
```diff
@@ -190,3 +191,94 @@ You might even be able to get some feedback early which could save you some time
...
+var Opal = asciidoctor.Opal;
+Opal.load('nodejs');
+Opal.load('pathname');
+require('asciidoctor-template.js');
+
+// Convert the document 'presentation.adoc' using Reveal.js backend
+var attributes = 'revealjsdir=node_modules/asciidoctor-reveal.js/node_modules/reveal.js';
+var options = Opal.hash({safe: 'safe',
+                         backend: 'revealjs',
+                         attributes: attributes});
```
obilodeau: Passing attributes as arguments makes overriding them in the document impossible. See the documentation on attribute assignment precedence for more information about that.

I suggest that you push the revealjsdir attribute default value in the example document as I suggested earlier.

obilodeau:
HACKING and README are too similar and HACKING is actually more complete. I'm merging this and I'll fix the redundancy by moving HACKING's content into README and pushing `revealjsdir` as a document attribute.

Thanks for the work!

HACKING.adoc
Mogztter: Actually I don't think that it's a good practice to define "rendering attribute" in the document. `revealjsdir` only make sense if my backend is `revealjs`.
My document could be rendered as an HTML5 presentation with Reveal.js or Bespoke.js or even as a PDF...

Mogztter:
> HACKING and README are too similar and HACKING is actually more complete. I'm merging this and I'll fix the redundancy by moving HACKING's content into README and pushing revealjsdir as a document attribute.
Thanks for the work!

Ok thanks 😄 
Do you have an account on http://npmjs.com/ ? Could you please publish a version ?

HACKING.adoc
obilodeau: I agree that a presentation _could_ be targeting several back-end but in 10+ presentations I did in asciidoc there was always something back-end specific. On the other hand, the reveal.js package location issue is real. Yours is not at the same location than mine. What do you propose as a solution to satisfy both location that don't require crippling OS packages? Maybe a check in the javascript shim? I would accept that.

obilodeau:
> Do you have an account on http://npmjs.com/ ?

I don't. Is it easy to take over the maintenance of a package? If so, I would suggest that you publish it under your name since you proposed to be the maintainer of the javascript side of this back-end.

Mogztter:
@obilodeau Ok, I don't remember how the "ownership" of a Node package works but I will try 👍

Mogztter:
> we should definitely get that template converter into the Asciidoctor organization. I've opened up transfer rights. You can either bring in that repository or merge it into asciidoctor.js. It's up to you how you want to manage it.

@mojavelinux Done! https://github.com/asciidoctor/asciidoctor-template.js

mojavelinux:
> Are you sure about that ? I can't reproduce this behavior. When using both Node 4.2.6 and 0.12.0

I was using Node 4.5, so maybe it was a bug in that version. Either way, I think Node 4 should be avoided since it was the first version after the big merge. I can see still supporting 0.12.0 since a lot of people were stuck on it, but even the last version of Fedora is on Node 6 now.

Regardless, it's just something to be aware if people do report it.

mojavelinux:
> Confirmed, this is related to the npm version.

Aha. That makes sense.

mojavelinux:
> I may have to change the default value of revealjsdir because the latest version of npm uses a flat hierarchy and reveal.js is at the root of node_modules .

I think we should simply say "If your node_modules does not have a flat hierarchy, then please update your version of npm. That's an indication that the version of npm on your machine is outdated."

HACKING.adoc
mojavelinux: Just add a trailing @ to the value and the attribute can be overridden by the document. Problem solved ;)

mojavelinux:
> Done! https://github.com/asciidoctor/asciidoctor-template.js

:+1: 

Permission updated!

obilodeau:
Ok found something. Just add me and @mojavelinux as owner on the package
you published with your account. This way we can cover just in case you are
unavailable. What do you think?

https://docs.npmjs.com/cli/owner

If you want to go this way let me know and I'll create an account.

mojavelinux:
:+1: That's exactly the arrangement I have on other packages I help maintain. In my opinion, npmjs.com has a better ownership model than other repositories for this reason.

HACKING.adoc
obilodeau: > Just add a trailing @ to the value and the attribute can be overridden by the document. 

And I was the one to refer to the documentation... :facepalm:

I'm opening up a PR with just that.

obilodeau:
My [new] npmjs.com username is obilodeau.

Mogztter:
@obilodeau @mojavelinux Package published! You are both collaborators :tada: 
We may need to use Markdown compatible syntax in README.adoc to improve the rendering :sob:

obilodeau:
Great! Are you planning to do an announcement on discuss.asciidoctor.org? I think downstream would benefit, namely, the AsciidocFX people (@rahmanusta) and adoc-editor (@mgreau) would benefit from this work. And there's the chrome extensions, etc. Getting the word out would help.

> We may need to use Markdown compatible syntax in README.adoc to improve the rendering :sob:

I googled a bit and didn't find an easy way out... I hoped that we could override the README.adoc with some field in `packages.json` (something similar can be done in Pypi) but didn't find anything... A really minimal README for the npmjs.com website that points to the github doc would satisfy me. The problem I see is if we put a README.md, it'll override our README.adoc as the default README which I don't want.

Mogztter:
Yes I will do an announcement but first I need to fix a major issue in the
Asciidoctor template.js to make it work on browsers.

I also want to improve the documentation and test the integration myself
before go "public".
We are still at an early stage and I don't want users to be disappointed
because this is kinda working.

Le 30 sept. 2016 5:49 AM, "Olivier Bilodeau" <notifications@github.com> a
écrit :

obilodeau:
Understood. Please note that I modified the structure of the documentation
in the pull request that turns this back-end into a ruby gem. To make it
consistent with the ruby doc. This is pending a review by @mojavelinux.

We also created a starter repository with bundler instructions. We might
want to do that for node also?

https://github.com/obilodeau/presentation-revealjs-starter

mojavelinux:
> I hoped that we could override the README.adoc with some field in packages.json (something similar can be done in Pypi) but didn't find anything

Yes, that is possible. That's what we do for the bespoke plugins. You can see an example here: https://github.com/opendevise/bespoke-multimedia. What I do is transform the AsciiDoc to Markdown (abbreviated) and add it the readme field in package.json. That's one approach.

The approach is to use the Markdown compatibility support in Asciidoctor to get a reasonable interpretation of the AsciiDoc. See  http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#markdown-compatibility.

Ultimately, we need to get npm to support AsciiDoc. Frankly, it's ridiculous that they don't.

obilodeau:
> Yes, that is possible. [...] You can see an example here [...]

PR incoming!!

> Ultimately, we need to get npm to support AsciiDoc. Frankly, it's ridiculous that they don't.

I filed an issue here: https://github.com/npm/npm/issues/14159. They use this parser for README's: https://www.npmjs.com/package/marky-markdown

I feel like this won't get fixed since it looks like a deep architectural problem (enabling mutliple README processing engine) and npm is clearly winning right now in terms of {mind,market}share. Maybe I'm wrong but the readme field interim is good enough for most projects to at least point to GitHub without having to sacrifice their own markup choice.

That said, I did the right thing :tm: by opening an issue so I feel good about that :wink:

Issue #96, opened 2016-09-16
Image background with vertical slides


kubamarchwicki:
Hi, 
The following Asciidoc

```
= Test slide deck
:imagesdir: images
:backend: revealjs

[%notitle]
== Opening

image::cover.jpg[background, size=cover]

== I'm just title, no background

=== First vertical with background

image::70s.jpg[canvas, size=cover]
```

outputs html

```
<section class="stack future" aria-hidden="true" style="top: 0px; display: block;" hidden="">
  <section style="top: 227px; display: block;" id="i_m_just_title_no_background" data-background-image="images/70s.jpg" data-background-size="cover"> <!--1-->
    <h2>I’m just title, no background</h2>
  </section>
  <section style="top: 330px; display: none;" aria-hidden="true" class="future" id="first_vertical_with_background" data-background-image="images/70s.jpg" data-background-size="cover">
    <h2>First vertical with background</h2>
  </section>
</section>
```

The first subsection has a `data-background-image` set - which I think is obsolete. 
in general the outer section section element has the first image as background - which I consider an error

To work this around you can place whatever empty image as outer section background, but that's rather hack than solution.

mojavelinux:
At first glance, it looks like the new logic to promote a block image to a background isn't considering nested sections correctly.

kubamarchwicki:
That's my idea as well, but I couldn't hack sections template to have it
working :(
Thus only an error report

-- Kuba

obilodeau:
Provided a simple test case :ballot_box_with_check: 

Testing this, be right back!

obilodeau:
Pushed a fix for it in #97.

obilodeau:
Merged #97 which fixes this issue.

PR #97, opened 2016-09-16
Fixed vertical slides and background problem

obilodeau:
As reported in #96, a first vertical slide w/o a background should not show the next vertical slide's background.

Here's a fix for it. I'm not 100% satisfied with it (given that it spans for longer than 80 chars) but it works.

Integrated the user-provided test case (and we love those! :+1:).

### DIFF
--- a/templates/slim/section.html.slim
+++ b/templates/slim/section.html.slim
@@ -13,8 +13,10 @@
 - data_background_image, data_background_size, data_background_repeat,
   data_background_transition = nil
 
+/ find all images under current node with attribute background or canvas
+/ as long as they are of the same level as we are currently processing
 - slide_images = find_by(:context => :image) { |image|
--   image.attr(1) == 'background' || image.attr(1) == 'canvas'
+-   (image.attr(1) == 'background' || image.attr(1) == 'canvas') && image.level == level
 - }
 - unless slide_images.nil? or slide_images.length == 0
   - slide_images.each do |image|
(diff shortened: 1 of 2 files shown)


templates/slim/section.html.slim
```diff
@@ -13,8 +13,10 @@
 - data_background_image, data_background_size, data_background_repeat,
   data_background_transition = nil
 
+/ find all images under current node with attribute background or canvas
+/ as long as they are of the same level as we are currently processing
 - slide_images = find_by(:context => :image) { |image|
--   image.attr(1) == 'background' || image.attr(1) == 'canvas'
+-   (image.attr(1) == 'background' || image.attr(1) == 'canvas') && image.level == level
```
mojavelinux: I don't recommend relying on the `level` attribute. That's been a source of confusion and I'm not convinced we've named it correctly.

Instead, I would check that the parent (or parent section) is self. This does make me realize that we need a depth restriction in find_by.

mojavelinux: It might also help if we had a parent_section attribute on block.

mojavelinux: For now, I recommend just creating a helper method.

obilodeau: Since I can't re-open a PR, I feel kind of lazy and would wait to have an actual problem (and bug report / test case) because I'm going to sleep and I'll forget (since it's in the closed queue)...

mojavelinux: I sent a PR. https://github.com/asciidoctor/asciidoctor-reveal.js/pull/100

Issue #70, opened 2016-04-27
Compliance : Subtitle partitioning


GrmpCerber:
[Subtitle partitioning](http://asciidoctor.org/docs/user-manual/#subtitle-partitioning) is not enforced on the first slide with the reveal.js backend.

GrmpCerber:
Tracked it to 5b59a3a but I don't know how to fix it

mojavelinux:
What do you expect to happen?

GrmpCerber:
Well, _according to the user manual*_, the title should be splitted in two (from the last colon or any given separator) the first part being the main title, the second part being a subtitle.

In my case the document is titled like this : 

    = Catch phrase: Longer more explicit title introducing the real concept

Which ends up with a cumbersome first slide since the whole line is in `h1`.
I would expect that the subtitle would be placed underneath the title in a smaller font.

Yet, as I was writing the current answer I re-read the user manual and noticed the note stating that : 

> As of Asciidoctor 1.5.2, subtitle partitioning is not implemented in the HTML 5 backend. 

*So I'm not sure anymore if this remark is legit

obilodeau:
I implemented preamble support not that long ago. So something like this
would work:

```
= title

Subtitle

== first non title slide

Content
```

obilodeau:
Since the html5 back-end does not do subtitle partitioning and that we support preamble which would achieve the same effect, I will close that issue. Feel free to re-open if you feel otherwise.

```
= title

Subtitle

== first nontitle slide

Content
```

mojavelinux:
> > As of Asciidoctor 1.5.2, subtitle partitioning is not implemented in the HTML 5 backend.
> *So I'm not sure anymore if this remark is legit

Actually, it is. It's still not implemented in the HTML5 backend. But this isn't the HTML5 backend...so we need not worry about it. There's still an API for it.

I have implemented subtitle partitioning in the Bespoke converter. I think the same should be done for the Reveal.js converter. See https://github.com/asciidoctor/asciidoctor-bespoke/blob/master/templates/slim/slide_title.html.slim#L11-L16.

mojavelinux:
Basically, the title should be split into h1 and h2, and perhaps wrapped in a div to make it possible to style the group.

PR #98, opened 2016-09-18
Implemented subtitle partitioning

obilodeau:
See #70. Removed unnecessary HTML tags if no authors are provided too.


mojavelinux:
Looks good. I noticed that you brought in a couple of functions you aren't yet using. Future use?

obilodeau:
Hahaha no, it's a mistake. I'll remove it.

obilodeau:
Done and applied author and preamble CSS classes to the relevant sections as a bonus.

mojavelinux:
:+1: 

I don't think you need to say where the constants and functions come from. That's useful for the PR but probably not long-term. However, you could say at the top that this is similar to the Bespoke converter so that developers get the cue to study that converter as a reference (and vice versa).

PR #99, opened 2016-09-22
Changed regex, so it does not remove first letter of id

JacobAae:
The regex removes the first character of the title in the id, making

 == P2P Applications

have "2p-applications" as id, failing in navigation, as it starts with a number.


mojavelinux:
Thanks @JacobAae! That is definitely the regexp we want.

@obilodeau this looks correct to me (in fact, it's very critical).

JacobAae:
With the Gradle/Jruby plugin, it broke the navigation. The solution is based on this issue: https://github.com/hakimel/reveal.js/issues/1230

obilodeau:
Tested and you are correct. Behavior was broken.

PR #100, opened 2016-09-23
find the background image block in a portable and reliable way

mojavelinux:


obilodeau:
Looks fine and no regression found with our [manual] test files. :+1:

PR #101, opened 2016-09-26
Fix issue with empty json & use instead empty string.

cmoulliard:
- Fix issue with empty json & use instead empty string
- Resolve https://github.com/asciidoctor/asciidoctor-reveal.js/issues/88


cmoulliard:
Can somebody apply this patch please as I would like to work on issue 104 which is base on the document.jade

mojavelinux:
@obilodeau This is perfectly safe to merge. I recommend proceeding.

PR #106, opened 2016-09-29
node: Using @ attribute prefix to lower CLI priority

obilodeau:
Related to discussion in #95. Allows our example to avoid :revealjsdir: document attribute in recent node versions while still allowing per-document overrides.

Took the opportunity to reverse the default value as discussed.

Let me know if it's good for you.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -63,8 +63,10 @@ Opal.load('pathname');
 require('asciidoctor-template.js');
 
 // Convert the document 'presentation.adoc' using Reveal.js backend
+var attributes = 'revealjsdir@=node_modules/reveal.js';
 var options = Opal.hash({safe: 'safe',
-                         backend: 'revealjs'});
+                         backend: 'revealjs',
+                         attributes: attributes});
 Asciidoctor.$convert_file('presentation.adoc', options); // <1>
 ----
 <1> Creates a file named `presentation.html` (in the directory where command is run)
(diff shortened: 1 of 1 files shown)


obilodeau:
No news is good news

mojavelinux:
Oops, that's the wrong spot for the @. It needs to go at the end of the attribute value, not the end of the attribute name. Sorry for not catching that sooner.

obilodeau:
I tested before submitting the PR and it worked.

Munging the value of an attribute (which is often user-provided) to remove a terminating @ and change behavior sounds risky to me. What if the attribute needs to finish with a @ ?

Anyway, I'm not questioning the asciidoctor feature but more reflecting on why I expected it to be on the attribute name. And moreso that it worked... I'll retest

mojavelinux:
The document override worked because you are essentially setting a different (unrelated) attribute.

As far as the rules for how to make an attribute writable by the document, that predated Asciidoctor. If the attribute needs to finish with an @, then you must use a character reference (e.g., &#64;)...though of course there are still edge cases.

Mogztter:
I'm confused with the new review system... I made a comment a few days ago but my review was pending.... :angry:

mojavelinux:
I always use "Add Single Comment" because I have yet to figure out how the new system works. I'd be happy to be enlightened though.

obilodeau:
You must "start a review" make your comments and then "finish the review" which will submit all the comments at once. This avoids races which is nice and lets you complete your review before any back and forth starts to happen. That said, the blog entry introducing the new feature and surrounding docs is really shy on details. I had to discover it by trying.

Another thing it adds is that you can say that your review is blocking the merge or not.

obilodeau:
> The document override worked because you are essentially setting a different (unrelated) attribute.

You were right. It worked because I had a reveal.js/ installed in the test repository which is the fallback when `revealjsdir` is not set. Committing the fix.

mojavelinux:
> Another thing it adds is that you can say that your review is blocking the merge or not.

That's probably the most interesting part.

One thing I've noticed is that there are certain comments I can't comment on. I guess these are review threads that have been closed...but I find it kind of annoying because it locks you out on adding additional information.

Clearly, I need to read about this feature ;)

PR #107, opened 2016-10-02
Fix warning: 'missing space before text'

ggrossetie:
Explicitly start a new text-block.

Reference: http://stackoverflow.com/questions/22181813/jade-missing-space-before-text-for-line-x-of-jade-file


obilodeau:
:+1:

PR #108, opened 2016-10-03
Added a minimal README for the npm package

obilodeau:
The idea is to provide a minimal README so that it points to our GitHub README and we don't need to update it too often.

Filed upstream issue: https://github.com/npm/npm/issues/14159

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -3,6 +3,7 @@
 :toclevels: 2
 :uri-nodejs-download: https://nodejs.org/en/download/
 
+// Changes made here should be sync'ed in packages.json's readme field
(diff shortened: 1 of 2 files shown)


obilodeau:
cc @mojavelinux and @Mogztter via #95

mojavelinux:
:+1: 

We are planning to create some sort of tool that automatically generates / syncs this field from the AsciiDoc. It's not a complex thing to do and it really feels like the best approach. For now, adding it manually gets the job done.

obilodeau:
Next upstream issue: https://github.com/npm/www/issues/42

PR #111, opened 2016-10-09
Prepare release 1.5.5-3

ggrossetie:
I can't use `1.0.0` anymore so I stick with the convention `1.5.5-{release-version}`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "asciidoctor-reveal.js",
-  "version": "1.5.5-2",
+  "version": "1.5.5-3",
   "description": "Reveal.js backend templates for Asciidoctor.js, implemented in Jade",
   "engines": {
     "node": ">=0.12"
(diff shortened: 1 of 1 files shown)


obilodeau:
Why not 1.0.0? I know you released 1.5.5 on npmjs but you can probably pull that back and replace it with 1.0.0 and the number of people affected will be small. This is not `left-pad` ;)

Mogztter:
As far as I know this is not possible to remove a version on npmjs.
left-pad was remove entierly and npmjs took some actions to prevent this
from happening again.

Le 9 oct. 2016 14:30, "Olivier Bilodeau" <notifications@github.com> a
écrit :

obilodeau:
I would, by far, prefer that we are in sync regarding version number and manage one changelog and release stream. It seems feasible: https://docs.npmjs.com/cli/unpublish

obilodeau:
I've [released v1.0.1](https://github.com/asciidoctor/asciidoctor-reveal.js/releases/tag/v1.0.1) which aligned the version on both packages. I've also modified the release process documentation in `HACKING.adoc` so that we release both packages at the same time for every release.

If you feel this is the wrong way to proceed, please let me know why.

Note that I didn't need to unpublish 1.5.5-2 and that 1.0.1 is still considered the latest version. I haven't dig to understand why.

Mogztter:
Thanks Olivier.
As far as I understand 1.5.5-3 is a "pre-release" so 1.0.1 is the first
stable version.

Le 13 oct. 2016 03:48, "Olivier Bilodeau" <notifications@github.com> a
écrit :

PR #115, opened 2016-10-23
Compile to JavaScript

ggrossetie:
- Compile the Reveal.js backend library (written in Ruby) to JavaScript using Opal
- Upgrade Asciidoctor.js to 1.5.5-3 and use the new wrapper API
- Use npm >= 3.x (with this version the transitive dependency `asciidoctor-template.js` is automatically fetched and therefore we can remove this step from the install procedure)
- Version `1.0.2.dev` is invalid in npm (replaced by `1.0.2-dev`)


obilodeau:
Nice work! I'll adjust the `-dev` instead of `.dev` to the ruby version as well so that they are in sync.

obilodeau:
I still haven't tested it (including build instructions) on my machine yet. Will do tomorrow.

Also, @mojavelinux what do you think?

obilodeau:
Added two minimal changes. Tested both node and ruby packages: works as expected.

Keeping this open a few days to give @mojavelinux the chance to comment but otherwise, it's a :+1: from me.

Mogztter:
Thanks @obilodeau :wink:

Mogztter:
Thanks :tada:

Issue #92, opened 2016-09-06
Implement asciidoctor-doctest testing


obilodeau:
We should implement [asciidoctor-doctest](https://github.com/asciidoctor/asciidoctor-doctest/). See [asciidoctor-backends](https://github.com/asciidoctor/asciidoctor-backends/) for an example implementation.

Idea / suggestion came from @jirutka during this discussion #63.
- [x] migrate from Makefile and manual tests to asciidoctor-doctest
- [x] create travis-ci integration and run doctest as part of build

obilodeau:
As mentioned by @mojavelinux, we should use rake for tests

jirutka:
Rake for tests? o.O Rake is like Make…

obilodeau:
To run the tests, instead of make

obilodeau:
See https://github.com/asciidoctor/asciidoctor-reveal.js/pull/110#issuecomment-251581555

obilodeau:
Faced a problem trying to get the tests to run, filed a bug to upstream project in order to get help: asciidoctor/asciidoctor-doctest#11

For anyone interested, work-in-progress branch is here: [obilodeau/asciidoctor-reveal.js](https://github.com/obilodeau/asciidoctor-reveal.js/tree/doctest)

It is, again, due to my misunderstanding of the whole gem, bundler and ruby packaging ecosystem...

obilodeau:
Making some progress. Until further notice, refer to the pull request #116 to follow what is happening.

obilodeau:
With #116 in master, this has been implemented.

PR #116, opened 2016-10-27
Asciidoctor-Doctest Testing

obilodeau:
The beginning of asciidoctor-doctest integration (#92).

At first, I thought this would help us test both slim and jade at the same time but I realize it would need a asciidoctor-doctest.js implementation. Maybe we could add a rake task to generate templates from the javascript/node toolchain and use doctest to do the diff. @Mogztter thoughts on this? 

Test samples were generated from current back-end so I think some thought, reviews and more custom examples should be migrated before calling it a go.

Also Travis integration would be really nice.

Tasks
- [x] port to using doctest 2.0.0.beta.3
- [x] travis-ci integration (rake test)
- [x] migrate at least one manual test into a non-embedded test
- [x] see what happens with asciidoctor/asciidoctor-doctest#12
- [x] review the currently generated HTML and consider dropping the irrelevant stuff

Comments and further contributions welcome!


Mogztter:
> At first, I thought this would help us test both slim and jade at the same time but I realize it would need a asciidoctor-doctest.js implementation. Maybe we could add a rake task to generate templates from the javascript/node toolchain and use doctest to do the diff. @Mogztter thoughts on this?

How did you generate files in `output` (ie. where is the source files) ? `input` directory is empty, is this intended ?

obilodeau:
> How did you generate files in output (ie. where is the source files) ? input directory is empty, is this intended ?

Yes it is normal. The files in `output/` match tests in `asciidoctor-doctest`'s own test set: https://github.com/asciidoctor/asciidoctor-doctest/tree/master/data/examples/asciidoc. Eventually, I'll migrate the manual tests over to `input/` which will need new matching results in `output/` but I haven't done so yet.

Mogztter:
Ok so I need to generate one file per example in JavaScript to mimic what Asciidoctor Doctest is doing ?
Can we copy `data/examples/asciidoc` files over ? It will be easier to read example files.

Currently `output` directory is what we get now when we are using Slim template or what the expected result should be ? Where should I put the result generated from Jade template ? Should we consider Slim template the reference ?

Thanks
Guillaume

mojavelinux:
The sooner we can get this merged, the better. Then we can all work on improving the examples. wdyt?

Mogztter:
I'm starting to upgrade Jade/Pug templates to use the new public API with Asciidoctor.js 1.5.6+. 

To make sure that I'm not breaking anything, I'm writing a test for each `.adoc` file in the `test/` directory. I'm guessing that maybe we should merge this pull request so I could use doctest in the JavaScrip tests suite ?

@obilodeau What's left to do ?

obilodeau:
I made a custom test working today. See `test/input/document.adoc`.

The manual tests are not ported to the doctest setup and I'm still trying to workaround the fact that all my tests would need to be in a very large `document.adoc`.

What would I need to do to test using the javascript toolchain?

obilodeau:
> [...] I'm still trying to workaround the fact that all my tests would need to be in a very large document.adoc.

Turned out this was an assumption I made. I am converting the tests to doctest right now (adding metadata). Making some reorganization while I'm at it. But now I need to get some :sleeping_bed:.

So all `test/*.adoc` files should have slim renders and be compared when running `bundle exec rake test` pretty soon. 

Hopefully, that rendering will be stable enough between jade and slim so that we can use two different templating engine but only one HTML result set. We'll see soon enough.

In the mean time, anything (instructions) that can help me run `rake tests` using the javascript toolchain would be appreciated :smile:

Mogztter:
I think it won't be easy to cross compile `asciidoctor-doctest` to JavaScript because the library has a few dependencies.
Maybe we could mimic the behavior of `doctest` in JavaScript to generate tests ?

The basic idea is to compare an input with an expected output:

```js
require('asciidoctor-template.js');
require('../build/asciidoctor-revealjs.js');
var assert = require('assert');
var fs = require('fs');

describe('Rendering', function () {

  var tests = [
    {input: './test/block_admonition.adoc', expected: './test/block_admonition_output.html'},
    {input: './test/block_audio.adoc', expected: './test/block_audio_output.html'},
    {input: '...', expected: '...'}
  ];

  tests.forEach(function (test) {
    it('should produce a valid Reveal.js presentation for ' + test.input, function () {
      var html = asciidoctor.convertFile(test.input, {safe: 'safe', backend: 'revealjs', to_file: false});
      assert.equal(html, fs.readFileSync(test.expected, 'utf-8'));
    });
  });
});
```

obilodeau:
> Maybe we could mimic the behavior of doctest in JavaScript to generate tests ?

The problem here is that we won't leverage doctest's power unless we reimplement it the same. Features like showing exactly where the problem is introduced with colorized diffs. Also, the output will probably be different since it will use another HTML beautifier and I'm afraid we won't be able to use the same test output for both slim and jade (ensuring compatibility between the two).

That said, I think this is the only real option we've got for now unless asciidoctor-doctest's maintainer is interested in making it transcompilable.

Mogztter:
> The problem here is that we won't leverage doctest's power unless we reimplement it the same. Features like showing exactly where the problem is introduced with colorized diffs. Also, the output will probably be different since it will use another HTML beautifier and I'm afraid we won't be able to use the same test output for both slim and jade (ensuring compatibility between the two).

That's right, you will have the diff output of the chosen JavaScript framework.
But the output is not that bad (at least with `mocha`):

mocha-html-output

obilodeau:
> So all test/*.adoc files should have slim renders and be compared when running bundle exec rake test pretty soon.

That's done. I also moved the test cases to `examples/` as I find it easier to discover for users. Using the doctest toolchain, nothing would prevent us to have both `examples/*.adoc` files for complete examples and `test/*.adoc` for corner-case or regression tests.

Because of asciidoctor/asciidoctor-doctest#12 most tests are useless now.

Next up is travis-ci integration (running `rake test` on github pushes)

jirutka:
@obilodeau https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116#issue-185566202:
> it would need a asciidoctor-doctest.js implementation

No, you don’t need. Doctest doesn’t need any deep integration with Asciidoctor, it can be modified to just call external program (asciidoctor.js) that accepts arguments and dumps output to stdout.

@Mogztter https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116#issuecomment-256930478:
> Ok so I need to generate one file per example in JavaScript to mimic what Asciidoctor Doctest is doing ?

No, one file may contain multiple examples, typically relevant to one particular AST node (e.g. `inline_anchor`, `olist`, …). It’s all described [here](https://github.com/asciidoctor/asciidoctor-doctest/#examples).

@Mogztter https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116#issuecomment-256930478:
> Can we copy data/examples/asciidoc files over ? It will be easier to read example files.

You can, but really shouldn’t. It’d be just code duplication with all its drawbacks.

@obilodeau https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116#issuecomment-323500568:
> I'm still trying to workaround the fact that all my tests would need to be in a very large document.adoc.

Why? This is not how it should be used.

@obilodeau https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116#issuecomment-323507198:
> Hopefully, that rendering will be stable enough between jade and slim so that we can use two different templating engine but only one HTML result set. We'll see soon enough.

Doctest “normalizes” HTML for comparison, so if both templates produce syntactically equal output (formatting, attributes order etc. is irrelevant), then they will be equal.

Why you need two sets of templates in two different templating engines? Asciidoctor.js is Asciidoctor transpiled to JS using Opal, right? Then why not to transpile even templates, specifically Ruby code generated by the template engine?

You can use [asciidoctor-templates-compiler](https://github.com/jirutka/asciidoctor-templates-compiler) to compile set of Slim templates into single Ruby file – Asciidoctor converter. If you avoid using Slim’s splat attributes (and maybe some other extra extensions), then the resulting code is standalone, i.e. doesn’t depend on Slim or Temple at all. I wrote it for [asciidoctor-html5s](https://github.com/jirutka/asciidoctor-html5s).

@Mogztter https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116#issuecomment-323520552:
> I think it won't be easy to cross compile asciidoctor-doctest to JavaScript because the library has a few dependencies.

Why you would do that? That’s not needed at all.

@obilodeau https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116#issuecomment-323527704:
> That said, I think this is the only real option we've got for now unless asciidoctor-doctest's maintainer is interested in making it transcompilable.

I’m not gonna make doctest transcompilable, because it doesn’t make any sense for me. As I know, Asciidoctor.js runs on node.js, so it should be possible to call it externally as a command, pass it some arguments and gather HTML on stdout, shouldn’t be? That’s all what’s needed for integration with Doctest. It currently does not support this out-of-box, but I’d be happy to implement it.

jirutka:
BTW, please use asciidoctor-doctest 2.0.0.beta.3 or [dev](https://github.com/asciidoctor/asciidoctor-doctest/tree/dev) branch, not the last stable or master branch. I’ve started almost complete rewrite of doctest two years ago and came back to it few weeks ago. The new version is much easier to integrate into project and also more cleanly extensible.

obilodeau:
Thanks @jirutka for the detailed answers.

> Why do you need two sets of templates in two different templating engines? Asciidoctor.js is Asciidoctor transpiled to JS using Opal, right? Then why not to transpile even templates, specifically Ruby code generated by the template engine?
>
> You can use asciidoctor-templates-compiler to compile set of Slim templates into single Ruby file – Asciidoctor converter. When you avoid using Slim’s splat attributes (and maybe some other extra extensions), then the resulting code is standalone, i.e. doesn’t depend on Slim or Temple at all. I wrote it for asciidoctor-html5s.

I think this right here is what we should evaluate first.

> As I know, Asciidoctor.js runs on node.js, so it should be possible to call it externally as a command, pass it some arguments and gather HTML on stdout, shouldn’t be? That’s all what’s needed for integration with Doctest. It currently does not support this out-of-box, but I’d be happy to implement it.

This would be an acceptable compromise if we can't get compiled templates to work.

@mojavelinux, @Mogztter: thoughts?

> BTW, please use asciidoctor-doctest 2.0.0.beta.3 or dev branch, not the last stable or master branch. I’ve started almost complete rewrite of doctest two years ago and came back to it few weeks ago. The new version is much easier to integrate into project and also more cleanly extensible.

Ok, I will update this PR so that we use 2.0.0.beta3.

Mogztter:
> I think this right here is what we should evaluate first.

Yes, Ruby converters (without dependency) are really easy to integrate in Asciidoctor.js (and it will avoid the code duplication between Slim and Jade/Pug templates).
The next question is, is it really necessary to use Slim templates ? Couldn't we just write the Reveal.js converter in Ruby (like the [default html5 converter](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb)) ?

> This would be an acceptable compromise if we can't get compiled templates to work.

Yes that could work.
We can create a simple Node.js program to run Asciidoctor.js or even use https://github.com/asciidoctor/asciidoctor-cli.js

> Why you would do that? That’s not needed at all.

I didn't know that doctest could be easily modified to call an external program, that's why 😉

jirutka:
> The next question is, is it really necessary to use Slim templates ? Couldn't we just write the Reveal.js converter in Ruby (like the [default html5 converter](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb)) ?

I think that the [default html5 converter](https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/converter/html5.rb) is a good example why it’s not a very good idea… Moreover, you already have Slim templates.

That said, I was thinking about writing the converter in plain Ruby, using some HTML builder. However, I didn’t find many promising HTML builders. The most interesting is probably [htmless](https://github.com/pitr-ch/htmless), but I find too complex, with slow initialization and probably memory hungry. Then I started thinking about writing simple HTML builder and realized that if I want somewhat elegant DSL, then it’d not be very fast or it’d be too complex. So I reconsidered Slim, learned how it works (and so learned about Temple), how it meets my goals and realized that it’s actually the best option.

> …or even use https://github.com/asciidoctor/asciidoctor-cli.js

Yes, this should be sufficient.

> I didn't know that doctest could be easily modified to call an external program, that's why 😉

Eh, what have you think? That it’s written using hieroglyphs hewn into stone and there’s no why how to understand it and modify? Have even looked into the code or read docs? If you at least grepped sources, then you’d find that there’s only one place where the code interacts with `Asciidoctor` and it’s only calling `Asciidoctor#convert`, i.e. pure string (+ options) in, string out. Obviously it can be modified to call an external program, even if it’d require patching the source (which is not).

mojavelinux:
> Eh, what did you think? That it’s written using hieroglyphs hewn into stone and there’s no way how to understand it and modify? Have even looked into the code or read docs?

I don't like your tone here. It's not respectful. You can make this point without talking down to your fellow community member. A simple "yes it can" would have sufficed.

jirutka:
> A simple "yes it can" would have sufficed.

Actually not, because that’s not the point. I feel a bit offended that he assumed that it’s not possible, apparently without even trying. I see two explanation. He thought that I produce totally unreadable and non-extensible code so it’d be easier to just reimplement it without even bothering to look at it and/or understand how it works. The second explanation would be really non-respectful

obilodeau:
> Eh, what did you think? That it’s written using hieroglyphs hewn into stone and there’s no way how to understand it and modify? Have even looked into the code or read docs?

Of course, source code can be read, understood and modified but the original goal of using doctest is saving us work by re-using yours. I extensively grepped the code yesterday, print-debugged the regexes, filed issues on doctest, which you resolved, you brought us good insights, which might save us even more work! So everything is great!

There is a ton of code out there that I use every day yet haven't read, please don't be offended if I didn't figure this out. Ruby and Javascript are like my 6th / 8th languages and I only write in these outside of work as a hobby for `asciidoctor-reveal.js` (the goal being: write my talks in asciidoc). I need guidance and skilled community members like you, @mojavelinux and @Mogztter to assist me in these complex ecosystems (node, ruby, template engines, asciidoctor, whatnot). Thanks again for your continued support and patience!

jirutka:
@obilodeau, my response was for @Mogztter, not for you! You’re approach is fully reasonable and I know that you’ve been debugging it and reported issues (and the behaviour with `:header_footer:` was really confusing, I did it wrong in this case).

mojavelinux:
> I think that the default html5 converter is a good example why it’s not a very good idea...Moreover, you already have Slim templates.

Although we don't agree on why the default converter is the way it is (I have my reasons), we do agree that this style should not be propagated to external converters. That's why I advocated for using Slim for the Reveal.js converter, and also used it for the Bespoke converter.

The drawback of that decision so far has been that it's been a challenge to transpile it to JavaScript. But I really like this idea that we can compile Slim to Ruby (with temple) and transpile it to JavaScript from there. I'd be okay in working around any restrictions in Slim. I'm very interested in this endeavor.

mojavelinux:
> I only write in these outside of work as a hobby for asciidoctor-reveal.js (the goal being: write my talks in asciidoc).

I appreciate all the time and effort you dedicate to this ecosystem. I want you to enjoy the time you spend on this hobby. The currency in open source is mutual respect and a shared sense of camaraderie. In other words, we pay each other in words. So we should always make sure it's a worthy offering as we exchange ideas and, yes, critiques.

mojavelinux:
> > Can we copy data/examples/asciidoc files over ? It will be easier to read example files.

> You can, but really shouldn’t. It’d be just code duplication with all its drawbacks.

The only difficulty I've had with doctest is that I wish the inputs for the tests were closer at hand. But I think that can be done without code duplication. The tool could simply copy them to a build folder before running the tests (and run them from there). It would be clear to the user they are just read-only copies, but they would be easier to locate in order to study them.

I'm also thinking that the datasets should be a separate project or repo. While the built-in tests are definitely helpful for converters that are designed to process the common input, there are other converters that have very different testing needs and therefore wouldn't need to test those built-in scenarios. If we had these datasets in a dedicated repository, then we could really perceive them as something that's providing a kind of spec for core converters while not being coupled (or not perceived to be coupled) with doctest for those using it for other use cases. (For example, I'd want completely original inputs for the Bespoke converter).

(we can take up that discussion elsewhere if you think it's worth persuing).

obilodeau:
More progress thanks to @jirutka's help

* DONE: port to using doctest 2.0.0.beta.3
* DONE: see what happens with asciidoctor/asciidoctor-doctest#12
* DONE: review the currently generated HTML and consider dropping the irrelevant stuff

Right now 10 tests are failing due to asciidoctor/asciidoctor-doctest#14.

Next up is Travis integration.

obilodeau:
First stab at Travis done. I started by targeting the same platforms as asciidoctor itself.

Aside from the expected above-mentioned 10 failures, I've got these ones too:

* ruby 2.0.0p648: nokogiri 1.8 requires ruby > 2.1 so there's a dependency failure here
* ruby 1.9.3: `There was a NameError while loading asciidoctor-doctest.gemspec`
* ruby 1.8.7: `[!] There was an error while loading 'asciidoctor-revealjs.gemspec': Malformed version number string 1.0.3-dev. Bundler cannot continue.`
* JRuby 9.0.5: -doctest failures filed here: https://github.com/asciidoctor/asciidoctor-doctest/issues/15
* JRuby 19mode and 18mode have the same problems as old ruby versions

I'm not sure what should I do with these. Is it acceptable to support only `>= 2.1` and JRuby `>= 9.1.12`? @mojavelinux, what do you think? Personally, I'm on Arch Linux so I run 2.4 and I'm quite disconnected with what's out there.

mojavelinux:
It is perfectly acceptable for extensions to only support versions of Ruby
that are still active as stated on ruby-lang.org or JRuby.org. That means
Ruby >= 2.2 and JRuby 9k. You might consider 2.1 too since that's still the
default on a lot of OSX installations afaik.

(I'm going to make the same change in core in the 1.6.0 release, though
that's a separate discussion).

obilodeau:
Travis-CI integration is done and since @jirutka fixed asciidoctor/asciidoctor-doctest#14. We are completely cleared! :smiley:

@jirutka: One last thing I would like you to do is to tag a release and push it to rubygems so that I can remove [this hack](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/116/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R3) and depend on `asciidoctor-doctest` in my gemspec as a development requirement instead.

obilodeau:
This PR is now ready for review. Thanks everyone for your help and patience!

HACKING.adoc
```diff
@@ -254,3 +249,33 @@ Then proceed as documented in the `README.adoc`.
...
+
+    bundle exec rake generate
+
+=== Custom tests
+
+Files in the `examples/` directory are used as tests.
+Resulting slides are kept in `test/output/slim/`.
+
+// TODO :header_footer: tests are broken.
+// See upstream issue: https://github.com/asciidoctor/asciidoctor-doctest/issues/12
```
jirutka: I guess that this TODO is already resolved. ;)

HACKING.adoc
```diff
@@ -254,3 +249,33 @@ Then proceed as documented in the `README.adoc`.
...
+However, if you have all dependencies properly installed this command should run the tests successfully:
+
+    rake test
+
+=== Generating HTML test target
+
+Tests were bootstrapped by generating them from asciidoctor-doctest's test corpus and current asciidoctor-revealjs' slim template engine.
+This is done using the following command:
+
+    bundle exec rake generate
```
jirutka: Maybe add note about `FORCE=y` for updating changed examples and add link to https://github.com/asciidoctor/asciidoctor-doctest/#generate-examples ?

asciidoctor-revealjs.gemspec
```diff
@@ -29,4 +29,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'thread_safe', '~> 0.3.5'
 
   s.add_development_dependency 'rake', '~> 10.4.2'
+  s.add_development_dependency 'asciidoctor-doctest', '~> 2.0.0.beta.4'
```
jirutka: I’d be better to fix this version (`= 2.0.0.beta.4`), ’cause it’s a beta and there _may_ come some incompatible changes.

test/output/slim/block_audio.html
```diff
@@ -0,0 +1,28 @@
+<!-- .basic -->
+<div class="audioblock">
+  <div class="content">
+    <audio controls="" src="ocean_waves.mp3">Your browser does not support the audio tag.</audio>
```
jirutka: There should be `controls`, without `=""` (and the same situation with `loop` and some other attributes). This is problem with nokogiri, it doesn’t not support HTML5. :( However, you can manually fix it in generated examples, it can parse it well and assert passes. It’s not a big deal though.

test/output/slim/block_admonition.html
```diff
@@ -0,0 +1,72 @@
...
+    <tr>
+      <td class="icon">
+        <div class="title">Warning</div>
+      </td>
+      <td class="content">Watch out for dragons!</td>
+    </tr>
+  </table>
+</div>
+
+<!-- .basic_multiline -->
```
jirutka: This file is outdated, I changed names format in beta.3 to use hyphen instead of underscore.

jirutka: You can use the following to update them:

```sh
sed -Ei '/^<!-- \.(\w+) -->/ s/_/-/g' *.html
sed -Ei '/^<!-- \.(\w+)$/ s/_/-/g' *.html
```

jirutka: Ah, and also I’ve dropped `block_` prefix in beta.3, to be consistent with Asciidoctor (IIRC `block_` prefix in node names is a legacy).

test/output/slim/document.html
```diff
@@ -0,0 +1,2194 @@
+<!-- .title -->
```
jirutka: Document examples should be scoped (using `:include:` option), look at https://github.com/jirutka/asciidoctor-html5s/blob/master/test/examples/html5/document.html for inspiration.

test/output/slim/inline_anchor.html
```diff
@@ -0,0 +1,43 @@
...
+
+<!-- .bibref-with-text -->
+
+<!-- .link -->
+
+<!-- .link-with-text -->
+
+<!-- .link-with-target-blank -->
+
+<!-- .link-with-role -->
```
jirutka: These empty examples don’t look good… some error in reveal.js converter…?

jirutka: Hm, it seems that you’ve somehow mixed examples generated using older DocTest version and the latest. It’d be better/easier to remove all output examples generated from reference examples and generate them from scratch.

asciidoctor-revealjs.gemspec
obilodeau: I'm a cybersecurity person and I think it's always better to enable upstream projects to patch their issues but I see your point. I'll pin to that version but I hope you'll keep us in the loop for improvements / security issues :wink:

test/output/slim/block_audio.html
obilodeau: File no longer exists after the refresh

obilodeau: Aaah it's named `audio.html` now. I fixed it. Commit coming soon.

test/output/slim/document.html
obilodeau: I see now why the options were coming from the output document! Because I have no control on the input ones. Makes sense. Thanks for the pointer!

test/output/slim/inline_anchor.html
obilodeau: Just done that

obilodeau: However, the `inline_anchor.html` template you mentioned just got re-generated the same way.

obilodeau:
@jirutka: I addressed all your comments.

Rakefile
```diff
@@ -0,0 +1,15 @@
...
+require 'tilt'
+
+DocTest::RakeTasks.new do |t|
+  t.output_examples :html, path: 'test/output/slim'
+  t.input_examples :asciidoc, path: [ *DocTest.examples_path, 'examples' ]
+  t.converter = DocTest::HTML::Converter
+  t.converter_opts = { template_dirs: 'templates/slim' }
+end
+
+# TODO add a rake task to render fully working examples to look at them in a browser
```
Mogztter: We should maybe create an issue to not forget to do it 😉

obilodeau: Don't you always grep for `TODO|FIXME` before you do a release? ;)

Mogztter: No, do you ? I grep from time to time but it's not integrated in my development process.
That's why I like to create issues on GitHub so I can keep track of what needs to be done.

jirutka:
There’s a lot of skipped examples (i.e. missing output examples), is it okay…? Run `bundle exec rake doctest:test VERBOSE=yes` to see which.

obilodeau:
I think skipped examples are fine for now. We can open a compliance issue about them but I would put them out of scope for what we are trying to achieve here.

Rakefile
```diff
@@ -0,0 +1,24 @@
...
+
+desc 'Renders all the test slides into fully working examples that you can look in a browser'
+# rendered slides will be put in examples/ directory
+task :render do
+#    Asciidoctor.convert_file 'data-background-newstyle.adoc',
+#        :backend => 'revealjs',
+#        :base_dir => 'examples'
+##        :template_dir => 'templates/slim'
+	system "bundle exec asciidoctor-revealjs data-background-oldstyle.adoc", :chdir => 'examples'
+end
```
obilodeau: @Mogztter: I didn't manage to make `convert_file` work. `base_dir` was not honored. Any ideas?

Mogztter: I'm guessing that you may have to configure the  `safe` mode.

Mogztter: The following seems to work:

```ruby
Asciidoctor.convert_file 'data-background-newstyle.adoc',
        :safe => 'safe',
        :backend => 'revealjs',
        :base_dir => 'examples'
```

obilodeau: Still not working for me:
```
$ bundle exec rake render 
rake aborted!
Errno::ENOENT: No such file or directory @ rb_sysopen - data-background-newstyle.adoc
/home/olivier/src/asciidoc/asciidoctor-reveal.js/.bundle/gems/ruby/2.4.0/gems/asciidoctor-1.5.6.1/lib/asciidoctor.rb:1575:in `initialize'
/home/olivier/src/asciidoc/asciidoctor-reveal.js/.bundle/gems/ruby/2.4.0/gems/asciidoctor-1.5.6.1/lib/asciidoctor.rb:1575:in `open'
/home/olivier/src/asciidoc/asciidoctor-reveal.js/.bundle/gems/ruby/2.4.0/gems/asciidoctor-1.5.6.1/lib/asciidoctor.rb:1575:in `convert_file'
/home/olivier/src/asciidoc/asciidoctor-reveal.js/Rakefile:19:in `block in <top (required)>'
/usr/bin/bundle:22:in `load'
/usr/bin/bundle:22:in `<main>'
Tasks: TOP => render
(See full trace by running task with --trace)
```

Should I change `cwd` with Rake? Anyway, I want the output to be in `examples/` and gitignore `examples/reveal.js/`.

Mogztter: Uh oh I see... I didn't run the right Rake task 😳 
What about:
```ruby
Asciidoctor.convert_file 'examples/data-background-newstyle.adoc',
        :safe => 'safe',
        :backend => 'revealjs',
        :base_dir => 'examples'
```

`to_dir` should be relative to `base_dir` so the document should be rendered in the `examples/` directory.

Mogztter: If you want the document to be rendered in `examples/reveal.js/`, I think you need to define `:to_dir => 'reveal.js'`.

obilodeau:
@jirutka I filed #134 for the skipped tests. I'm merging this now.

Thanks everyone for your patience and help through this! It should help development and avoid regressions a lot! :smile:

jirutka:
Please next time rebase PR before merging, i.e. remove fixup commits and other noise to get more clean repo history.

PR #121, opened 2017-01-26
Documented admonition support

obilodeau:


obilodeau:
Grace period over. Merging myself since this is just doc and tests which rarely hurt.

PR #123, opened 2017-02-10
Apply data-state 'title' to the title slide

nicolaiparlog:
Some fancy stuff mind end up outside of slide sections and to give it special treatment on the title slide, the section's CSS class `title` does not suffice. (See a use case [here](http://slides.codefx.org/junit-5/2017-01-25-JUG-Karlsruhe), where the footer only shows from the second slide forward.)

This change applies the data-state `title` to the title slide, making reveal.js apply it as a CSS class to the document when the title slide is active.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -1,6 +1,7 @@
 = {project-name} Changelog
 :project-name: asciidoctor-reveal.js
 :uri-repo: https://github.com/asciidoctor/asciidoctor-reveal.js
+:uri-issue: {uri-repo}/issues/
 
 This document provides a high-level view of the changes introduced in {project-name} by release.
 For a detailed view of what has changed, refer to the {uri-repo}/commits/master[commit history] on GitHub.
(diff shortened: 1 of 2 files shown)


obilodeau:
Looks good. The slides too!

Is the source / [S]CSS available somewhere? I would like to see how you've done footnotes so I can document it in our README.

I'll manually test your PR and merge.

obilodeau:
Works as advertised and found [your repo](https://github.com/CodeFX-org/slides). I'll try to submit a proposal for footers in asciidoctor-reveal.js based on your stuff I think.

nicolaiparlog:
Oh, sorry for not reacting, was afk. Good thing, you found the repo.

> I'll try to submit a proposal for footers in asciidoctor-reveal.js based on your stuff I think.

That would be awesome! :smiley: Note that I use almost (but not quite) the same content on the title slide. Maybe you could provide a nice way to reuse the same content there.

obilodeau:
> Note that I use almost (but not quite) the same content on the title slide. Maybe you could provide a nice way to reuse the same content there.

Covering all cases is not going to be easy. I invite you to subscribe to notifications in #53 and evaluate what we come up with.

PR #126, opened 2017-03-29
resolves #125 coerce Reveal.js settings to boolean values

### LINKED ISSUE #125: null / "" are not mapped to false / true correctly?
Having an issue (asciidoctor/asciidoctor-maven-plugin#288) with building slides from asciidoc with the Maven plugin, @mojavelinux pointed out that the plugin mapping of boolean values is correct but values are not mapped correctly by the reveal converter:

E.g., setting
```xml
<revealjs_controls>true</revealjs_controls>
```
leads to
```js
Reveal.initialize({
  // Display controls in the bottom right corner
  controls: ,
```
making JS complain about the , instead of an expected value of true or false.

This results in an empty slide show.

mojavelinux:


obilodeau:
Thanks for taking the time!

PR #129, opened 2017-08-14
Update asciidoctor-revealjs.gemspec

PierreR:
Is it possible to update the gemset to the latest release version of asciidoctor ?

I would like to add `asciidoctor-revealjs` to the list of gems in [nixos](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/typesetting/asciidoctor/Gemfile) without carrying different versions of asciidoctor.


mojavelinux:
Technically, 1.5.6.1 is already covered by the version in the gemspec, since it's a fuzzy match.

There's no harm in showing 1.5.6 for intent, though.

PR #131, opened 2017-08-24
Compiling slim templates into pure ruby (simplifies Javascript maintenance)

obilodeau:
I'm leveraging asciidoctor-templates-compiler by @jirutka. The goal here is to get rid of the jade templates and only have to maintain one set of templates. See #63.

~~This is not ready, I want doctest (#116) merged in first to see for potential regressions using doctest.
Will rebase/fix once it is in.~~

- [x] ruby side proof of concept
- [x] adapt to make it work from javascript
- [x] wait for #116 to go in
- [x] integrate doctest properly: see https://github.com/jirutka/asciidoctor-html5s/blob/master/Rakefile#L53
- [x] doctest test for rendering regressions / improvements
- [x] verify ruby-side packaging scripts/docs
- [x] verify javascript-side packaging scripts/docs
- [ ] wait on the next asciidoctor.js release that integrate this fix: https://github.com/asciidoctor/asciidoctor.js/pull/370
- [ ] ~re-enable ruby 2.1, 2.2 travis support by compiling templates using a recent version of ruby but using it from an old one~
- [x] re-enable ruby 2.1, 2.2 travis support by using asciidoctor-templates-compiler 0.3.0

Waiting on upstream

- [x] Remove `replaceUnsupportedFeatures` function from [npm/builder.js](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/131/files#diff-32d7599354463e657c27d56ccc4b16bf) once judofyr/temple#112 is merged and released
- [x] Remove `capture_generator` from [lib/asciidoctor-templates-compiler.rb](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/131/files#diff-afc31ec40a96a0effaf59190d1fce685R7) once judofyr/temple#113 is merged and released (or at least fix the typo `capture_generate` -> `capture_generator`)

Thanks to @jirutka for the asciidoctor-templates-compiler! It took me only an hour maybe to get the PoC running.


Mogztter:
> adapt to make it work from javascript

@obilodeau You can remove the following lines:
https://github.com/asciidoctor/asciidoctor-reveal.js/blob/a1fddf4da604fdfce3b7d569d1bf4076132e451a/lib/asciidoctor-revealjs.rb#L3
https://github.com/asciidoctor/asciidoctor-reveal.js/blob/a1fddf4da604fdfce3b7d569d1bf4076132e451a/package.json#L43

Since `asciidoctor-template.js` is not required anymore.

templates/slim/helpers.rb
```diff
@@ -3,14 +3,13 @@
 require 'asciidoctor'
```
Mogztter: This does not work in Asciidoctor.js because Opal will try to load this module from the `OPAL_PATH`.

templates/slim/helpers.rb
```diff
@@ -3,14 +3,13 @@
...
 require 'json'
 
+# Needed only in compile-time.
+require 'slim-htag' if defined? Slim
+
 if Gem::Version.new(Asciidoctor::VERSION) <= Gem::Version.new('1.5.3')
```
Mogztter: Opal does not implement `Gem`. You can still check the version in JavaScript with:

```js
asciidoctor.getCoreVersion(); // available in the next release of Asciidoctor.js
```

Though you will need to use something like that to compare the version:

https://github.com/asciidoctor/asciidoctor.js/blob/34e64c3e95293a03d459421a55fa34978b971066/spec/share/asciidoctor.spec.js#L3-L7

Mogztter:
I'm making progress on my branch https://github.com/mogztter/asciidoctor-reveal.js/tree/compile-slim-templates

NOTE: You should use `asciidoctor.js` master branch.

Mutable String methods are not supported in Opal:

```
NotImplementedError: String#<< not supported. Mutable String methods are not supported in Opal.
```
In the generated converter, we are using `String#<<`. For instance:

```ruby
_buf = ''
if (has_role? 'aside') or (has_role? 'speaker')
  _buf << ("<aside class=\"notes\">".freeze)
  _buf << ((content).to_s)
  _buf << ("</aside>".freeze)
else
  _buf << ("<div".freeze)
  # ...
end
```

Could we use the same strategy as Asciidoctor core and use an Array instead ?

```ruby
_buf = []
if (has_role? 'aside') or (has_role? 'speaker')
  _buf << ("<aside class=\"notes\">".freeze)
  _buf << ((content).to_s)
  _buf << ("</aside>".freeze)
else
  _buf << ("<div".freeze)
  # ...
end
_buf * LF
```

I'm not sure about the performance impact but if @mojavelinux is doing it in Asciidoctor core I think this should be fine ?
As an alternative we could replace it in the JavaScript build in order to keep mutable String in Ruby.

@jirutka What do you think ?

Mogztter:
I've added a hack to replace mutable operations on String: https://github.com/Mogztter/asciidoctor-reveal.js/commit/ef168e06b06bdda4027714d54461f9cdf8004827#diff-32d7599354463e657c27d56ccc4b16bfR61

```js
// Replace variable = '' by variable = []
data = data.replace(/([^ ]*) = ''/g, '$1 = \[\]');

// Replace _buf by _buf * ''
data = data.replace(/_buf$/g, '_buf * \'\'');

// Replace variable (alone in a single line) by variable * ''
data = data.replace(/^(\s*)(_[a-z1-9_\[\]]*)$/gm, '$1$2 = $2 * \'\'');
```

For now I replace in-place the `converter.rb` file but we could generate a new file (dedicated to Asciidoctor.js)

Here's the result:
```
= Title

== Slide 1
Content 1

== Slide 2
Content 2
```
```
<section id="_slide_1"><h2>Slide 1</h2><div class="paragraph"><p>Content 1</p></div></section>
<section id="_slide_2"><h2>Slide 2</h2><div class="paragraph"><p>Content 2</p></div></section>
```

templates/slim/helpers.rb
jirutka: This check is not necessary, IMHO it should not be included in JS variant.

jirutka:
> Is the json module required ? By default the opal-compiler does not include this module but I can it if necessary.

I don’t know why it’s here, it doesn’t seem to be used in reveal.js templates.

> Could we use the same strategy as Asciidoctor core and use an Array instead ?

Yes, just use different Temple generator – `Temple::Generators::ArrayBuffer` instead of default `Temple::Generators::StringBuffer`.

There’s currently no configuration option for this in asciidoctor-templates-compiler (I’ll add it), but you can subclass or patch [Asciidoctor::TemplatesCompiler::Slim](https://github.com/jirutka/asciidoctor-templates-compiler/blob/master/lib/asciidoctor/templates_compiler/slim.rb) and modify `#engine_options`.

> I've added a hack to replace mutable operations on String…

Hacks are not needed here (see above), Slim and especially Temple (backend for code generation) is nicely configurable.

templates/slim/helpers.rb
```diff
@@ -3,14 +3,13 @@
 require 'asciidoctor'
 require 'json'
 
+# Needed only in compile-time.
+require 'slim-htag' if defined? Slim
```
jirutka: This is not required, I’ve added it to my `helpers.rb` just to allow using Slim templates even directly without ahead of time compilation.

templates/slim/helpers.rb
jirutka: It’s probably used only for the version check in this `helpers.rb`.

templates/slim/block_table.html.slim
```diff
@@ -24,9 +25,10 @@ table(id=@id class=['tableblock',"frame-#{attr :frame, 'all'}","grid-#{attr :gri
                   - cell_content = cell.text
                 - else
                   - cell_content = cell.content
-              *{:tag=>(tblsec == :head ? 'th' : 'td'), :class=>['tableblock',"halign-#{cell.attr :halign}","valign-#{cell.attr :valign}"],
+              = html_tag(tblsec == :head || cell.style == :header ? 'th' : 'td',
```
jirutka: `html_tag` is not provided by Asciidoctor, it’s defined in my `helpers.rb`:

https://github.com/jirutka/asciidoctor-html5s/blob/a71db48a1dd5196b668b3a3d93693c5d877c5bf3/data/templates/helpers.rb#L84-L108

I was thinking about creating Slim/Temple extension like [slim-htag](https://github.com/jirutka/slim-htag), but for arbitrary tag. Actually I’ve implemented it and it works, but I don’t know how to name it (and how to name the special attribute) and if it’s really a good thing to do: it’s not idiomatic and it does break The Principle of Least Surprise.

It may look like this:

```rb
??? [
  ??=(tblsec == :head || cell.style == :header ? 'th' : 'td')
  class=['tableblock', "halign-#{cell.attr :halign}", "valign-#{cell.attr :valign}"]
  colspan=cell.colspan
  rowspan=cell.rowspan
  style=((@document.attr? :cellbgcolor) ? %(background-color:#{@document.attr :cellbgcolor};) : nil) ]
```

`???` should be name of the special “tag” (e.g. `element`, `dyntag`, …?) and `??` name of the special attribute defining the actual tag (e.g. `tag`, `_`, …?).

Mogztter:
> Yes, just use different Temple generator – Temple::Generators::ArrayBuffer instead of default Temple::Generators::StringBuffer.

Super nice 🤓 

> Hacks are not needed here (see above), Slim and especially Temple (backend for code generation) is nicely configurable.

I didn't know about `Temple::Generators::ArrayBuffer`, thanks for sharing 👍 

> There’s currently no configuration option for this in asciidoctor-templates-compiler (I’ll add it), but you can subclass or patch Asciidoctor::TemplatesCompiler::Slim and modify #engine_options.

Ok I will remove my hack and give it a try.

Mogztter:
> Yes, just use different Temple generator – Temple::Generators::ArrayBuffer instead of default Temple::Generators::StringBuffer.

@jirutka It's almost working, the buffer is now an array but there's still some String variables. Do you think we can configure how these variables are generated ?

```ruby
_buf = []
_buf << ("<div".freeze) # OK
```

```ruby
_temple_html_attributeremover1 = ''
_temple_html_attributeremover1 << ((_temple_html_attributemerger1.reject(&:empty?).join(" ")).to_s) # KO
```

Mogztter:
> verify javascript-side packaging scripts/docs

@obilodeau As mentioned by @jirutka, we should remove the generated file `dist/main.js` from the repository. This is a common practice in JavaScript but a bad one. I think we should use the same approach as Asciidoctor.js. The `dist` files are generated by Travis when we need to publish a new release.

jirutka:
> It's almost working, the buffer is now an array but there's still some String variables.

It seems that you have to set `generator: Temple::Generators::ArrayBuffer` and also `capture_generator: 'ArrayBuffer'`, but I’m not sure, it’s quite strange.

templates/slim/helpers.rb
obilodeau: Is that version check even necessary given that the way to use this module is through bundler?

templates/slim/helpers.rb
obilodeau: Repeating my above comment: Is that version check even necessary given that the way to use this module is through bundler for ruby (which would take care of making sure proper version of asciidoctor is there)?

templates/slim/helpers.rb
obilodeau: For development it might be useful to avoid the build step. Worth keeping as currently scoped in the `unless opal` block. I'm very open to remove it if there's a good argument against though.

templates/slim/block_table.html.slim
obilodeau: After quickly looking at [HTML documentation](https://html.spec.whatwg.org/multipage/semantics.html#semantics) I would say `???` be `html_element` and `??` = `tag_name` but that's far from compact. `dyntag` and `tag` would work for me as shorter alternatives.

obilodeau: In the meantime, I've added your `html_tag` function to our `helpers.rb`

obilodeau:
@Mogztter I pulled in your changes and added some of mine on top. Ideally to avoid rebasing it would be nice if you would pull my tree (or start a branch from scratch) so that we keep the history clean.

Mogztter:
> It seems that you have to set generator: Temple::Generators::ArrayBuffer and also capture_generator: 'ArrayBuffer', but I’m not sure, it’s quite strange.

Indeed, we can configure `capture_generator` on the `Temple:Generators`:
https://github.com/judofyr/temple/blob/2dafd522d60acc7b912ec0da4756ed2bb0160d15/lib/temple/generator.rb#L12-L15

It's one step closer but the generated code still use `String#<<`:

**before.rb**
```ruby
_buf = []
_buf << ("<div".freeze)
_temple_html_attributeremover1 = ''
_temple_html_attributemerger1 = []
_temple_html_attributemerger1[0] = "paragraph"
_temple_html_attributemerger1[1] = ''

# ... KO!
_temple_html_attributeremover1 << ((_temple_html_attributemerger1.reject(&:empty?).join(" ")).to_s)
_temple_html_attributeremover1

# ... KO!
_temple_html_attributemerger1[1] << ((_slim_codeattributes1).to_s)
```

**after.rb**

```ruby
_buf = []
_buf << ("<div".freeze)
_temple_html_attributeremover1 = []
_temple_html_attributemerger1 = []
_temple_html_attributemerger1[0] = "paragraph"
_temple_html_attributemerger1[1] = ''

# ... OK
temple_html_attributeremover1 << (_temple_html_attributemerger1.reject(&:empty?).join(" "))
_temple_html_attributeremover1 = _temple_html_attributeremover1.join("".freeze)

# ... KO!
_temple_html_attributemerger1[1] << ((_slim_codeattributes1).to_s)
```

Mogztter:
@obilodeau I've updated my branch: https://github.com/Mogztter/asciidoctor-reveal.js/commits/compile-slim-templates

obilodeau:
@Mogztter: your changes have been pushed.

Mogztter:
Thanks!
I made a typo in the file `lib/asciidoctor-templates-compiler.rb`, it's `capture_generator` not `capture_generate`.

Also the build does not work on Ruby 2.1 and Ruby 2.2 since the library `asciidoctor-templates-compile` requires Ruby >= 2.3.
Should we remove Ruby 2.1 and Ruby 2.2 ?

Mogztter:
Also the build with `jruby-9k` does not work:

```
Bad Syntax
/home/travis/.rvm/gems/jruby-9.1.7.0/gems/ruby-beautify-0.97.4/lib/ruby-beautify.rb:20:in `pretty_string'
...
```

templates/slim/helpers.rb
jirutka: It’s not necessary, I suggest to remove it.

templates/slim/helpers.rb
jirutka: No, it’s not necessary. I suggest to remove it.

templates/slim/helpers.rb
jirutka: As you said, it might be useful in development, for someone.

jirutka:
> It's one step closer but the generated code still use String#<<:

Hm, this looks like a bug in Slim or Temple. I’ll take a look at it later.

> …since the library `asciidoctor-templates-compile` requires Ruby >= 2.3.

Yeah, because I use `<<~` (squiggly heredoc) operator introduced in 2.3. `asciidoctor-templates-compile` is needed only in build time, so it should not be a problem. You can use different Ruby version to compile the templates. Travis uses rvm, so it should be easy.

> Also the build with `jruby-9k` does not work:

That’s bad, ruby-beautify executes `ruby` as external process: https://github.com/erniebrodeur/ruby-beautify/blob/v0.97.4/lib/ruby-beautify.rb#L54-L58. It should be at least parametrized…

You can avoid this error by running only fast compilation (`pretty: false`).

templates/slim/helpers.rb
```diff
@@ -31,6 +31,38 @@ def to_boolean val
     val && val != 'false' && val.to_s != '0' || false
   end
 
+  ##
+  # This function is from the asciidictor-html5s project
+  # Copyright 2014-2017 Jakub Jirutka <jakub@jirutka.cz> and the Asciidoctor Project.
+  # Licensed under the MIT License.
```
jirutka: It’s not needed to copy copyright here; both projects are MIT licensed and contain Asciidoctor project as one of the copyright holders.

templates/slim/helpers.rb
```diff
@@ -31,6 +31,38 @@ def to_boolean val
...
+  def html_tag(name, attributes = {}, content = nil)
+    attrs = attributes.inject([]) do |attrs, (k, v)|
+      next attrs if !v || v.nil_or_empty?
+      v = v.compact.join(' ') if v.is_a? Array
+      attrs << (v == true ? k : %(#{k}="#{v}"))
+    end
+    attrs_str = attrs.empty? ? '' : attrs.join(' ').prepend(' ')
+
+
+    if VOID_ELEMENTS.include? name.to_s
```
jirutka: Where’s `VOID_ELEMENTS` constant…?

jirutka:
>> It's one step closer but the generated code still use String#<<:
>
> Hm, this looks like a bug in Slim or Temple. I’ll take a look at it later.

It’s a bug in Temple, I’ve fixed it and opened PR https://github.com/judofyr/temple/pull/112.

>> Also the build with jruby-9k does not work:
>
> That’s bad, ruby-beautify executes `ruby` as external process…

I’ve fixed it, opened PR https://github.com/erniebrodeur/ruby-beautify/pull/43 and patched it in `asciidoctor-templates-compiler` 0.1.3. It’s now tested against JRuby on CI.

>> It seems that you have to set generator: Temple::Generators::ArrayBuffer and also capture_generator: 'ArrayBuffer', but I’m not sure, it’s quite strange.

I’ve opened PR https://github.com/judofyr/temple/pull/113 to change default `capture_generator` to more reasonable value.

Mogztter:
Awesome! Thanks @jirutka 💯 
Hopefully your pull requests will be merged quickly.

Mogztter:
**TODO**

 - [ ] Remove Ruby `2.1` and `2.2` from [.travis.yml](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/131/files#diff-354f30a63fb0907d4ad57269548329e3)
 - [ ] Remove `replaceUnsupportedFeatures` function from [npm/builder.js](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/131/files#diff-32d7599354463e657c27d56ccc4b16bf) once judofyr/temple#112 is merged and released
 - [ ] Remove `capture_generator` from [lib/asciidoctor-templates-compiler.rb](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/131/files#diff-afc31ec40a96a0effaf59190d1fce685R7) once judofyr/temple#113 is merged and released (or at least fix the typo `capture_generate` -> `capture_generator`)

templates/slim/helpers.rb
obilodeau: Fixed

templates/slim/helpers.rb
obilodeau: Left behind. I added it.

obilodeau:
Merged with master (which now has doctest testing #116). Addressed several comments.

Doctest are failing, I think it is all related to the handling of `<h level=X>` instead of `<hX>` (which I've never heard of before). Note: this is still using asciidoctor 1.5.4 (see #132). I'll port to 1.5.6.1 in a different PR.

I'll try to fix doctest failures. It's just too late now to continue working on this.

>  You can use different Ruby version to compile the templates. Travis uses rvm, so it should be easy.

I could use a hand on this.

jirutka:
> Doctest are failing, I think it is all related to the handling of <h level=X> instead of <hX> (which I've never heard of before).

This means that [slim-htag](https://github.com/jirutka/slim-htag) was not loaded. `<h level=X>` is not a real element, it should be converted to `<hX>`.
It is loaded in https://github.com/asciidoctor/asciidoctor-reveal.js/pull/131/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR18 and even https://github.com/asciidoctor/asciidoctor-reveal.js/pull/131/files#diff-a327a0ed9729c22f5d32cb5593a372aeR7, so I don’t understand why it doesn’t work for you.

obilodeau:
> This means that slim-htag was not loaded. <h level=X> is not a real element, it should be converted to <hX>.

doctests were not using the ruby back-end but the slim templates directly. Fixed it.

jirutka:
> doctests were not using the ruby back-end but the slim templates directly.

Hm, still it should work; [templates/slim/helpers.rb:7](https://github.com/obilodeau/asciidoctor-reveal.js/blob/adbba16910795ae5869f5ad4ffb9ef5faae861f7/templates/slim/helpers.rb#L7) does not do the job for some reason. :/

obilodeau:
All Travis checks are now green.

Should we merge this and open tickets for remaining tasks or is there a minimal javascript-side test that you would like performed @Mogztter?

Others, it is also a good time to speak. For me, this is a go-in pending a documentation update (in HACKING mostly) and a rebase before entering review time.

Mogztter:
I think we can iterate on this feature.
Now that the converter is a single Ruby file, tested with Doctest, we will add JavaScript-side tests as soon as https://github.com/asciidoctor/asciidoctor-doctest/issues/13 is resolved

jirutka:
Plz don't merge yet.

jirutka:
FYI, if you want to fallback to built-in converter for nodes you don’t have template for, set `delegate_backend: 'html5'` in asciidoctor-templates-compiler.

jirutka:
I’ve released asciidoctor-templates-compiler 0.2.0 with these important additions:

* new parameter `engine_opts` to pass options into Temple Engine,
* patch Temple to pass options to capture_generator,
* patch Temple to use the same type of generator for capture_generator by default.

So you can remove all the workarounds.

I found out that Opal does not handle `String::freeze` correctly, so you must disable it when generating converter for Opal: `Temple::Generators::ArrayBuffer.new(freeze_static: false)`. See [this revision of Rakefile](https://github.com/jirutka/asciidoctor-html5s/blob/687477719e1de4f7c93c6c75db90e9052389dfd4/Rakefile#L75) in my repo.

Rakefile
```diff
@@ -28,6 +33,10 @@ namespace :build do
             outfilesuffix: '.html',
             filetype: 'html',
           },
+	  delegate_backend: 'html5',
```
jirutka: Indentation… 😠

templates/slim/helpers.rb
```diff
@@ -5,10 +5,6 @@
 
   # Needed only in compile-time.
   require 'slim-htag' if defined? Slim
```
jirutka: You can remove this as well, it doesn’t work as intended anyway.

Rakefile
obilodeau: Fixed. My vim isn't properly setup for ruby still :disappointed:

templates/slim/helpers.rb
obilodeau: done

obilodeau:
Current status:

* from ruby: everything seems fine
* from javascript: I can't even run it, I followed `HACKING.adoc` doc to set it up (unpublished node module) but I think it needs adjustments now maybe?

```
$ node build.js 
/home/olivier/src/asciidoc/presentation-asciidoctorjs-test/build.js:3
var Asciidoctor = asciidoctor.Asciidoctor();
                              ^

TypeError: asciidoctor.Asciidoctor is not a function
    at Object.<anonymous> (/home/olivier/src/asciidoc/presentation-asciidoctorjs-test/build.js:3:31)
    at Module._compile (module.js:573:30)
    at Object.Module._extensions..js (module.js:584:10)
    at Module.load (module.js:507:32)
    at tryModuleLoad (module.js:470:12)
    at Function.Module._load (module.js:462:3)
    at Function.Module.runMain (module.js:609:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:598:3
```

build.js:
```javascript
// Load asciidoctor.js + asciidoctor-reveal.js
var Asciidoctor = asciidoctor.Asciidoctor();

require('asciidoctor-reveal.js');

// Convert the document 'presentation.adoc' using Reveal.js backend
var attributes = 'revealjsdir=node_modules/reveal.js@';
var options = {safe: 'safe', backend: 'revealjs', attributes: attributes};
Asciidoctor.convertFile('presentation.adoc', options);
```

package.json:
```javascript
{
  "name": "presentation-asciidoctorjs-test",
  "version": "1.0.0",
  "description": "",
  "main": "build.js",
  "dependencies": {
    "asciidoctor-reveal.js": "file:../asciidoctor-reveal.js",
    "asciidoctor.js": "^1.5.5-5"
  },
  "devDependencies": {},
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}
```

package-lock.json: https://gist.github.com/obilodeau/bfba282a43e5e9572ce1d4f7cdf06b69

@Mogztter any hints?

Mogztter:
You need to remove this line :

```js
var Asciidoctor = asciidoctor.Asciidoctor();
```

And use `asciidoctor` (with lowercase) here :

```js
asciidoctor.convertFile('presentation.adoc', options);
```

Also attributes can now be defined as JSON object:

```js
var attributes = { 'revealjsdir': 'node_modules/reveal.js@' };
```

Mogztter:
@obilodeau We also need to update `bestikk-opal-compiler` to `0.3.0-integration7` (based on Opal 0.11), if we want to be compatible with Asciidoctor.js 1.5.6. Though the latest version of Asciidoctor.js `1.5.6-preview.3` does not work because Opal cannot resolve `ThreadSafe.Cache`. This bug has been fixed on master https://github.com/asciidoctor/asciidoctor.js/pull/370 but I need to publish a new preview version.

obilodeau:
@Mogztter I pulled in your change and updated `package.json` according to your last comment.

As expected, I'm stuck at the `ThreadSafe.Cache` issue:
```
$ node asciidoctor-revealjs.js 

/home/olivier/src/asciidoc/presentation-asciidoctorjs-test/node_modules/asciidoctor.js/dist/asciidoctor.js:26264
              } else { throw $err; }
                       ^
Cache: uninitialized constant ThreadSafe::Cache
```

I'm going to add a task to wait until the fix is released before rebase/review time.

obilodeau:
> You can use different Ruby version to compile the templates. Travis uses rvm, so it should be easy.

@jirutka I tried this. Using two different approaches. I'm close but it still doesn't work...

Can you look at it and give me a pointer?

## Attempt 1

Build matrix with shell environment variable and I alter my script's behavior. 
https://github.com/obilodeau/asciidoctor-reveal.js/compare/compile-slim-templates...obilodeau:travis-build-attempts-shell

Remaining problems:

### on 2.2

it fails because it says it requires temple (althought it shouldn't) but the `bundle install --without development` acts kind of weird (failed on first attempt then retry works)
https://travis-ci.org/obilodeau/asciidoctor-reveal.js/jobs/272199895

### on 2.1

`bundle install --without development` fails, looks like the --without simply doesn't work
https://travis-ci.org/obilodeau/asciidoctor-reveal.js/jobs/272199903

## Attempt 2

Forcing ruby 2.4 on all `rake build` then back to original ruby.
https://github.com/obilodeau/asciidoctor-reveal.js/compare/compile-slim-templates...obilodeau:travis-build-attempts

Problems:

### 2.4, 2.3, jruby

rake missing: https://travis-ci.org/obilodeau/asciidoctor-reveal.js/jobs/272211812, jruby: https://travis-ci.org/obilodeau/asciidoctor-reveal.js/jobs/272211820

### 2.2

```
LoadError: cannot load such file -- asciidoctor/doctest
```
https://travis-ci.org/obilodeau/asciidoctor-reveal.js/jobs/272211816

### 2.1

```
cannot load such file -- asciidoctor/doctest
```
https://travis-ci.org/obilodeau/asciidoctor-reveal.js/jobs/272211818

I've spent way to much time on this than I'm willing to admit. Guidance would be appreciated.

jirutka:
There are more traps than I’ve expected, but it works: https://github.com/jirutka/asciidoctor-html5s/blob/try-old-ruby/.travis.yml and https://github.com/jirutka/asciidoctor-html5s/blob/try-old-ruby/Rakefile. I’ll reconsider adding support for older rubies.

jirutka:
Okay, I’ve just released asciidoctor-templates-compiler 0.3.0 that is compatible with Ruby 2.1+ (and also adds CLI).

obilodeau:
Tonight:
- [x] bump to a-t-c 0.3.0
- [x] documentation updates
- [x] rebase
- [x] little tree re-org and dropped jade templates
- [x] all targeted platforms are green

Will file these issues:
- [ ] wait on the next asciidoctor.js release that integrate this fix: asciidoctor/asciidoctor.js#370

Ready for review.

I won't release 1.1.0 until this is fixed but I think it can go into master and get more exposure now.

obilodeau:
I filed #143 and #144. Last chance for reviews before this goes in.

npm/builder.js
```diff
@@ -49,7 +51,18 @@ Builder.prototype.compile = function (callback) {
   var opalCompiler = new OpalCompiler({dynamicRequireLevel: 'ignore'});
   opalCompiler.compile('asciidoctor-revealjs', 'build/asciidoctor-revealjs.js', ['lib']);
   typeof callback === 'function' && callback();
-}
+};
+
+Builder.prototype.replaceUnsupportedFeatures = function (callback) {
```
Mogztter: Maybe add a comment to state that this is hack that need to be removed once issue XX is resolved ?

jirutka: This is already not needed, see my comments.

obilodeau: I just removed it and confirmed things reach the same `ThreadSafe::Cache` error without it. It was reintroduced in the rebase process. The rebase was quite complicated to merge since that branch started before doctest was in and master was pulled in at some point.

PR #135, opened 2017-09-04
resolves #132 fix templates to work with 1.5.6

### LINKED ISSUE #132: Asciidoctor 1.5.6 introduced a rendering regression for data-background new-style images
The update of asciidoctor to 1.5.6 (in #129), introduced a regression. The test `data-background-newstyle.adoc` no longer renders properly with Asciidoctor 1.5.6. Rolling back to 1.5.4 fixed it.

Found while implementing doctest support in #116.

Plan:

I'm going to fix master, pin asciidoctor to 1.5.4 and do a release (w/o doctest or #131). This way people pulling 1.0.3 or current latest will have it working. As will people allowing bugfix updates only.

Then I'll fix this bug (hopefully fixable locally), increase dependencies to 1.5.6, merge #116, #131 and do a `1.1.0` release.

@mojavelinux, thoughts?

mojavelinux:
- use attributes[1] to access first positional attribute
- use textlabel to access textual label of admonition block

### DIFF
--- a/asciidoctor-revealjs.gemspec
+++ b/asciidoctor-revealjs.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
   s.extra_rdoc_files = Dir['README.adoc', 'LICENSE.adoc', 'HACKING.adoc']
   s.require_paths = ['lib']
 
-  s.add_runtime_dependency 'asciidoctor', '= 1.5.4'
+  s.add_runtime_dependency 'asciidoctor', '~> 1.5.6'
   s.add_runtime_dependency 'slim', '~> 3.0.6'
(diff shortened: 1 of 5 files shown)


obilodeau:
Looking good! Thanks!

I'm thinking of holding off a point release and wait for 1.1.0 to release this. I'm concerned about the current status of the jade templates and 1.1.0 should resolve these problems via #131 and we are almost there.

Your thoughts?

mojavelinux:
I think we can be patient for a release. Let's get this in master and allow others to road test it first.

mojavelinux:
(oh, it is in master...well, then let the road test begin!)

PR #136, opened 2017-09-04
Travis: building on asciidoctor's master too (related to #132)

obilodeau:
Thanks @mojavelinux to always point in the right direction :smile:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ rvm:
 - 2.2
 - 2.1
 - jruby-9k
+gemfile:
+- Gemfile
+- Gemfile.upstream
 env:
   global:
   # use system libraries to speed up installation of nokogiri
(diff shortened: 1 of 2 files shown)


mojavelinux:
This is great. I'll add it to the upstream trigger list.

Now, if you want to only run against upstream on the upstream triggered build, you can create an additional file named .travis-upstream-only.yml. The contents of that file get merged into the .travis.yml config, which can either override or update the settings.

Here's what that file would need to contain in your case:

```yml
matrix:
  exclude:
  - gemfile: Gemfile
```

We are asking it to modify the matrix to exclude the regular build.

See https://docs.travis-ci.com/user/customizing-the-build/#Excluding-Jobs for details.

mojavelinux:
Asciidoctor.js uses this. See https://github.com/asciidoctor/asciidoctor.js/blob/master/.travis-upstream-only.yml

Each project has it's own exclude rules of course.

mojavelinux:
It finally works!

https://travis-ci.org/asciidoctor/asciidoctor-reveal.js/builds/271898978

I had to do some tweaking of permissions.

PR #142, opened 2017-09-21
Fix adminition that had '|| @caption' appended

JacobAae:
When making slides and using the admonition block, a "|| @caption" was appended. This changes the parenthesis, to where I think it should be, adding the caprion only if there is no textlabel

image


mojavelinux:
This is definitely a bug, though I'm not sure the fix has the desired result. I think it needs to be:

```
i class=%(fa fa-#{icon_mapping[attr :name]}) title=((attr :textlabel) || @caption)
```

Slim is clearly getting confused about precedence.

Another way to write it is:

```
title=(attr :textlabel, @caption)
```

though that has slightly different behavior.

jirutka:
> ```slim
> i class=%(fa fa-#{icon_mapping[attr :name]}) title=(attr :textlabel) || @caption
> ```

Slim syntax is not so complicated, `|| @caption` is interpreted as a text content of the element (see [docs](https://github.com/slim-template/slim#text-content)), that should not be surprising.

As you’ve figured out, you must wrap it in parenthesis: `…title=((attr :textlabel) || @caption)` (or `…title=(attr :textlabel, @caption)`), or in idiomatic Ruby syntax: `…title=(attr(:textlabel) || @caption)` (or `…title=attr(:textlabel, @caption)`).
```

mojavelinux:
> Slim is clearly getting confused about precedence.

I say that only because it differs from Ruby rules. In Ruby, everything to the right of the equal sign is implicitly contained in round brackets.

To Slim's credit, we're not really talking about the same equals sign since this is an attribute list. So it should look more like an array entry, which would require the brackets.

obilodeau:
I've tracked the regression to this commit: https://github.com/asciidoctor/asciidoctor-reveal.js/commit/e6344890#diff-a550ba4d478c3d43dff573f7d35df4de.

I thought the PR was missing a fix (the other `|| @caption` introduced) so I generated test cases. Turns out this PR solves the only problematic case.

Before I merge it however, I would like #131 to go in since there's a folder re-org in there and I would have to rebase #131 again which I would rather avoid.

obilodeau:
Can you rebase your pull request please?

Basically, `templates/slim/block_admonition.html.slim` should now be `templates/admonition.html.slim`, `test/output/slim/admonitions-icons.html` should now be `test/doctest/admonitions-icons.html` and `test/output/slim/admonitions.html` should now be `test/doctest/admonitions.html`.

Thanks!

obilodeau:
:+1:

PR #146, opened 2017-09-26
Sync with Reveal.js 3.5.0

rahulsom:
There are a lot of things that've shown up in the reveal.js repo.
This brings in the missing configuration properties here.


obilodeau:
Please look at the build failures. They are related to your change.

obilodeau:
Just realized that the README would need adjustments as well.

obilodeau:
After some testing, I realize that the `defaultTiming` settings needs to be entirely optional, otherwise speaker notes always have an annoying per-slide timeout with red warnings.

I'll experiment with trying to avoid that problem later.

obilodeau:
Using null works so I'm doing that.

obilodeau:
Just rebased on master. Will test everything manually using both ruby and javascript stacks.

Issue #143, opened 2017-09-22
asciidoctor.js support is currently broken


obilodeau:
We need to wait for upstream to release a new version with this fixed integrated: asciidoctor/asciidoctor.js#370

This issue blocks the 1.1.0 release since we have no javascript story without asciidoctor.js

Stack trace:
```
$ node asciidoctor-revealjs.js 

/home/olivier/src/asciidoc/presentation-asciidoctorjs-test/node_modules/asciidoctor.js/dist/asciidoctor.js:26264
              } else { throw $err; }
                       ^
Cache: uninitialized constant ThreadSafe::Cache
```

mojavelinux:
We're almost ready for an Asciidoctor.js release. There's just one more issue to resolve. I'm going to try to get to it as soon as I can...this weekend looks probable.

obilodeau:
No rush. I just opened the ticket to remember and tick checkboxes on the #131 PR :wink:

mojavelinux:
FYI, I updated the PR that was holding back the release. We just need to settle another issue regarding the Opal upgrade and we should be good to go.

Mogztter:
@obilodeau How do you generate the file `lib/asciidoctor-revealjs/converter.rb` in "Opal mode" ? In `Rakefile` we are using this condition:

```ruby
  generator = if :mode == :opal
    Temple::Generators::ArrayBuffer.new(freeze_static: false)
  else
    Temple::Generators::StringBuffer
  end
```

I've tried different syntax, but I'm unable to set `mode => :opal`

jirutka:
This code snippet is wrong, `mode` should be a variable, not a symbol.

Mogztter:
@jirutka I've moved this block and replaced `:mode` by `args[:mode]`, see https://github.com/asciidoctor/asciidoctor-reveal.js/pull/156/files#diff-52c976fc38ed2b4e3b1192f8a8e24cffR21
Not sure if this was the original intention but now I can run this task with `Rake::Task[CONVERTER_FILE].invoke(:opal)`

jirutka:
Yes, but I eventually changed the approach for reasons I don’t remember right now, now I recommend this: https://github.com/jirutka/asciidoctor-html5s/blob/opal/Rakefile.

PR #149, opened 2017-09-28
Converting the examples using javascript / node toolchain

obilodeau:
Just a start of javascript testing. People with javascript and npm experience please comment. I borrowed this from Asciidoctor.js and Asciidoctor-template.js.

Tests fail exactly where it should. Waiting on #143 to be fixed.

### DIFF
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -67,7 +67,7 @@ Builder.prototype.copyToDist = function (callback) {
 Builder.prototype.removeDistDirSync = function () {
   log.debug('remove dist directory');
   bfs.removeSync('dist');
-  bfs.mkdirsSync('dist/templates');
+  bfs.mkdirsSync('dist');
 };
 
 Builder.prototype.examples = function (callback) {
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -101,8 +101,13 @@ Builder.prototype.convertExamples = function (callback) {
   fs.readdir(this.examplesDir, (err, files) => {
     files.forEach(filename => {
       if (path.extname(filename) == '.adoc') {
-        asciidoctor.convertFile(path.join(this.examplesDir, filename), options);
-        log.info(`Processed ${filename}`);
+        try {
+          asciidoctor.convertFile(path.join(this.examplesDir, filename), options);
+          log.info(`Successfully converted ${filename}`);
+        }
+        catch (err) {
+          log.error(`Error converting ${filename}: ${err}`);
+        }
       }
     });
   })
--- a/npm/builder.js
+++ b/npm/builder.js
@@ -3,6 +3,7 @@ module.exports = Builder;
 var async = require('async');
 var log = require('bestikk-log');
 var bfs = require('bestikk-fs');
+var fs = require('fs');
 var path = require('path');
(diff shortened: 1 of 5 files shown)


npm/builder.js
```diff
@@ -67,3 +69,34 @@ Builder.prototype.removeDistDirSync = function () {
...
+
+Builder.prototype.examples = function (callback) {
+  const builder = this;
+
+  async.series([
+    callback => builder.build(callback), // Build
+    callback => builder.compileExamples(callback), // Compile examples
+  ], () => {
+    log.info(`
+In order to visualize the result, a local HTTP server must be started within the root of this project otherwise you will have cross-origin issues.
```
Mogztter: I don't think it applies in this context

npm/builder.js
```diff
@@ -67,3 +69,34 @@ Builder.prototype.removeDistDirSync = function () {
...
+For this purpose, you can run the following command to start a HTTP server locally: 'npm run server'.`);
+    log.success(`You can now open: build/examples/`);
+    typeof callback === 'function' && callback();
+  });
+};
+
+Builder.prototype.compileExamples = function (callback) {
```
Mogztter: I think the naming is misleading since we are not really compiling examples

npm/examples.js
```diff
@@ -0,0 +1,6 @@
+const log = require('bestikk-log');
+const Builder = require('./builder.js');
+const builder = new Builder();
+
+log.task('examples');
+builder.examples();
```
Mogztter: As mentioned in https://github.com/asciidoctor/asciidoctor-reveal.js/pull/156#issuecomment-335613495 I think we should use an "examples" task to demonstrate how to use the converter in JavaScript.

In the Asciidoctor.js repository, I have one example with a textarea on the left of the page and the rendered Reveal.js presentation on the right of the page.

npm/builder.js
```diff
@@ -48,7 +50,7 @@ Builder.prototype.removeBuildDirSync = function () {
 Builder.prototype.compile = function (callback) {
   log.task('compile');
   var opalCompiler = new OpalCompiler({dynamicRequireLevel: 'ignore'});
-  opalCompiler.compile('asciidoctor-revealjs', 'build/asciidoctor-revealjs.js', ['lib']);
+  opalCompiler.compile('asciidoctor-revealjs', 'build/asciidoctor-reveal.js', ['lib']);
```
Mogztter: You are probably right the `revealjs.js` is a bit too much :wink:

npm/builder.js
obilodeau: Removed

npm/examples.js
obilodeau: Not sure I understand what you mean. We have it in the documentation already. Adding an example  would mean maintaining two copies of the same code and instructions.

npm/builder.js
obilodeau: Fixed. Changed to convert.

npm/examples.js
Mogztter: Alright, I thought the purpose of the examples task was to provide a "live" example.

Mogztter:
🎉 🎉 🎉

Issue #127, opened 2017-04-05
reveal.js history feature doesn't work with slides rendered by asciidoctor-reveal.js


obilodeau:
The setting `:revealjs_history: true` doesn't work.

Jumping directly to slides w/o a title is working with history enabled. These are replaced with a number.

What doesn't work is jumping directly to slides with a title.

To reproduce: 
* load this URL: http://gosecure.github.io/presentations/2016-12-01_botconf/malboxes.html#/_malboxes

Expected: that you arrived at the title slide "Malboxes"

Actual result: You land on the first slide of the deck.

Links to slide number works though: http://gosecure.github.io/presentations/2016-12-01_botconf/malboxes.html#/2

obilodeau:
Related to #99

obilodeau:
I have a fix ready. Expect PR soon.

obilodeau:
Fixed with the #150 merge.

PR #150, opened 2017-09-28
Reveal.js compatible slide id generation fixes history feature (#127)

obilodeau:
There were still some problems with the id stripping regex. I created a more extensive test case and made sure to pass all tests.

References:
* #127 
* #99 
* #75 
* hakimel/reveal.js#1346
* #32 
* https://github.com/hakimel/reveal.js/blob/360bc940062711db9b8020ce4e848f6c37014481/js/reveal.js

### DIFF
--- a/examples/history-regression-tests.adoc
+++ b/examples/history-regression-tests.adoc
@@ -14,4 +14,21 @@
 == P3rhaps this won't work
 // Second char is a number
 
+== 5th Slide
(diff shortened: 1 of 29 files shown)


obilodeau:
As per our policy for merges: since this is a clear bugfix with backing test cases, I'm going to merge this even if it got no real comments or reviews

PR #173, opened 2018-01-01
Fixed the display of include:: links for github and a broken link

obilodeau:
Broken link found while reviewing #169, a 1.1.0 regression.

@mojavelinux I would like your opinion on the way I workaround GitHub's *include:: are turned into link:* behavior.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -220,17 +220,17 @@ There are a few ways to have no titles on slides.
 * Adding the `notitle` option to your slide
 * Adding the `conceal` option to your slide
 
+ifeval::[{safe-mode-level} >= 20]
+See <<examples/concealed-slide-titles.adoc#,concealed-slide-titles.adoc>>.
+endif::[]
+ifeval::[{safe-mode-level} < 20]
 Here is an example of the three techniques in action:
 
-ifdef::env-github,env-browser[]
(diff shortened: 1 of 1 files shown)


mojavelinux:
What I recommend is just making a regular link when the GitHub environment is detected. No need to use an include since it won't get interpreted anyway.

```
ifeval::[{safe-mode-level} >= 20]
See <<examples/basic-example.adoc#,basic-example.adoc>>.
endif::[]
ifeval::[{safe-mode-level} < 20]
.basic-example.adoc
[source,asciidoc]
....
include::examples/basic-example.adoc[]
....
endif::[]
```

Technically it's better to check the safe mode level since that's what determines whether includes are enabled or not.

obilodeau:
I made the suggested changes.

One thing though, a link to an `adoc` file on GitHub will display its converted version not the source. I think this could be confusing to users. We could link to a `/raw/` link ([example](https://github.com/obilodeau/asciidoctor-reveal.js/raw/doc-fix-github-links/examples/title-slide-image.adoc)) but then the HTML version would no longer work offline as the link would no longer be relative.

Thoughts?

obilodeau:
I created #176 to have that discussion since it is not release-critical and we need to do a release $soon.

mojavelinux:
:+1:

PR #174, opened 2018-01-01
Adding converter.rb manually in gemspec. Fixes #172

### LINKED ISSUE #172: lib/asciidoctor-revealjs/converter.rb missing in 1.1.0
Today, I updated the gem from 1.0.4 to 1.1.0 . I get the following error now:

```
LoadError: cannot load such file -- .bundle/gems/ruby/2.4.0/gems/asciidoctor-revealjs-1.1.0/lib/asciidoctor-revealjs/converter
/.bundle/gems/ruby/2.4.0/gems/asciidoctor-revealjs-1.1.0/lib/asciidoctor-revealjs.rb:5:in `require_relative'
.bundle/gems/ruby/2.4.0/gems/asciidoctor-revealjs-1.1.0/lib/asciidoctor-revealjs.rb:5:in `<top (required)>'
```

Indeed, the file converter.rb is missing (it exists in 1.0.4 though).

Btw, something is also wrong with the official website: http://asciidoctor.org/docs/asciidoctor-revealjs/ (but I don't know, if this issue tracker is the right one - therefor, I didn't post an extra issue here).

obilodeau:
It's not the cleanest thing I've done in Ruby but it works. Reviewers, feel free to refer me to a solution better aligned with Asciidoctor's approach if you have one in mind.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -14,6 +14,9 @@ Enhancements::
 Compliance::
   * Code listing callouts now work properly ({uri-issue}166[#166], {uri-issue}167[#167])
 
+Bug fixes::
+  * The version 1.1.0 Ruby Gem was broken due to a packaging error ({uri-issue}172[#172])
+
 ////
 === Release meta
 
--- a/asciidoctor-revealjs.gemspec
+++ b/asciidoctor-revealjs.gemspec
@@ -14,7 +14,12 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 1.9.3'
 
   files = begin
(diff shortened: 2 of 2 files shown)


mojavelinux:
That's actually correct. Since you are building the file dynamically, you should need to list it explicitly in the gemspec. That shows the correct intent.

However, the correct place to add it would be inside the `files = begin` block since it's part of the aggregation, not the filtering.

```ruby
  files = begin
    if (result = Open3.popen3('git ls-files -z') {|_, out| out.read }.split %(\0)).empty?
      Dir['**/*']
    else
      result + ['lib/asciidoctor-revealjs/converter.rb']
    end
  rescue
    Dir['**/*']
  end
```

obilodeau:
Integrated your suggestion. Thanks!

mojavelinux:
Perfect!

Issue #118, opened 2016-12-17
Additional reveal.js dependencies/plugins


akhayyat:
How can I configure additional reveal.js plugins?

Reveal.js lists a number of plugins: https://github.com/hakimel/reveal.js/wiki/Plugins,-Tools-and-Hardware

For example, to use the reveal.js-menu plugin (https://github.com/denehyg/reveal.js-menu), I'd need at least to specify additional reveal.js dependences:

```js
Reveal.initialize({
    // ...
    dependencies: [
        // ... 
        { src: 'plugin/menu/menu.js' }
    ]
});
```
And at best, to be able to customize the plugin by specifying additional `Revial.initialize` attributes (see https://github.com/denehyg/reveal.js-menu#configuration).

Is there currently any way to specify additional reveal.js dependencies?

mojavelinux:
There doesn't seem to be a way currently. I guess that means we need to design a solution.

Perhaps a comma-separated list of extensions, like:

```
:revealjs_ext: menu,...
```

But that doesn't address the configuration for that plugin. That feels like too much for document attributes.

Of course, you can achieve this today simply by providing a custom document template. Though, perhaps that shines light on what the solution should be. We should read the dependencies block from a template (or json configuration file) that you can override. That way, you get full control.

Either way, we need to design a solution.

obilodeau:
>Perhaps a comma-separated list of extensions, like:
>
>     :revealjs_ext: menu,...

It's not an easy one to address since the plugin can be separately installed via bower, npm or manually which changes the way you need to write the `src:` entry.

> But that doesn't address the configuration for that plugin. That feels like too much for document attributes.

## Idea a) attribute to load external javascript file

Maybe we should allow for JSON to be injected via an external file. Quick testing shows that overriding keys would even work so we could provide default values and allow the user to override:

```javascript
{"a":"b","a":"c"}
> Object {a: "c"}
```

So a new attribute `revealjs_initialize:` used like:

    :revealjs_initialize: config.js

With a `config.js` file like:

```javascript
menu: {
        side: 'right',
        themes: false,
        transitions: true,
    },
```

## Idea b) attribute to run individual `Reveal.configure()` statements

Reveal.js supports updating it's configuration after the `initialize(...)` call with:

    Reveal.configure({ autoSlide: 0 });

As [documented here](https://github.com/hakimel/reveal.js#configuration).

Something like:

[...]

PR #196, opened 2018-06-10
Config plugins

a4z:
this is a suggestion to solve #118 through allowing to specify optional files that can be used to add and configure plugins.
Please see changes in README.adoc for more information

### DIFF
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -198,8 +198,13 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
             { src: '#{revealjsdir}/plugin/markdown/markdown.js', condition: function() { return !!document.querySelector( '[data-markdown]' ); } },
             #{(attr? 'source-highlighter', 'highlightjs') ? "{ src: '#{revealjsdir}/plugin/highlight/highlight.js', async: true, callback: function() { hljs.initHighlightingOnLoad(); } }," : nil}
             { src: '#{revealjsdir}/plugin/zoom-js/zoom.js', async: true },
-            { src: '#{revealjsdir}/plugin/notes/notes.js', async: true }
-        ]
+            { src: '#{revealjsdir}/plugin/notes/notes.js', async: true },
+            { src: '#{revealjsdir}/plugin/print-pdf/print-pdf.js', async: true } ,
+            #{(attr? 'revealjs_plugins_add') ? File.read(attr('revealjs_plugins_add', '')) : ""}
+        ], 
+
+        #{(attr? 'revealjs_plugins_setup') ? File.read(attr('revealjs_plugins_setup', '')) : ""}
+
       });
     - unless (docinfo_content = (docinfo :footer, '.html')).empty?
       =docinfo_content
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -21,7 +21,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
       link rel='stylesheet' href='#{revealjsdir}/css/theme/#{attr 'revealjs_theme', 'black'}.css' id='theme'
     - if attr? :icons, 'font'
       - if attr? 'iconfont-remote'
-        link rel='stylesheet' href=(attr 'iconfont-cdn', %(#{cdn_base}/font-awesome/4.3.0/css/font-awesome.min.css))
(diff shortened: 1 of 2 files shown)


a4z:
no idea why or what on travis if failing. changes are working great here. please help!

obilodeau:
Thanks for your contribution! Sorry I haven't replied earlier. A quick glance and everything looks fine. Plus, I like the design. I'm going to test it later and report back.

Travis failures are related to the fact that your code changed the output of the generated HTML. We do strict regression testing against these. Don't worry about it, I can fix that.

One thing I'm wondering: I'm not sure attributes is a list, I think it's a hashmap (key / value) so multiple  `revealjs_plugins_add` statements will overwrite each other. I'll verify that by trying to add two plugins.

a4z:
Thanks for your feedback!
multiple revealjs_plugins_add makes no sense, the file file given needs to have a comma separated list of file path to plugins

example from what I use:

cat plugins_add.js
```
{ src: 'config/plugins/mouse-pointer.js' },
{ src: 'config/plugins/badges.js' }
```

cat plugins_setup.js
```
       badges: {
          languages: true
        }
```

if the concept of this solution will be approved I can improve the documentation and put up a sample 


knowing that it is just about some (maybe withe spaces) in a test seems fixable, but I am on holidays this week, can have a look at it in 2 weeks if you have not time

templates/document.html.slim
```diff
@@ -21,7 +21,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
       link rel='stylesheet' href='#{revealjsdir}/css/theme/#{attr 'revealjs_theme', 'black'}.css' id='theme'
     - if attr? :icons, 'font'
       - if attr? 'iconfont-remote'
-        link rel='stylesheet' href=(attr 'iconfont-cdn', %(#{cdn_base}/font-awesome/4.3.0/css/font-awesome.min.css))
+        link rel='stylesheet' href=(attr 'iconfont-cdn', %(#{cdn_base}/font-awesome/4.5.0/css/font-awesome.min.css))
```
obilodeau: This should be done in a separate PR

a4z:
I am glad you like the solution! 
Thanks for you effort to fix the open points and your work to pull this in!

obilodeau:
Made the discussed changes. Feel free to comment. Will merge if no objections.

Thanks for your contribution @a4z!

a4z:
Glad you like it!
The only thing I noticed since last time, and where I started to think about how to solve it, is that the notes plugin (where you see the next slide in a new window) does not work via revealjs over CDN.
In this case the node plugin needs to be added manual to a custom folder.
So maybe it should be removed from the default loaded plugins.
Also, the loaded markdown plugins may not be used ever, and loading them should be avoided.
So I start to think if there should be any plugin loaded per default....

obilodeau:
I use speaker notes extensively and I never noticed. I never used reveal.js hosted by a CDN because I don't trust conference networks :wink:. It's probably blocked because of a same origin policy violation. I know speaker notes relies on PostMessage to pass state around. Are you sure hosting only the notes plugin locally fixes the issue?

Also, if you want reveal.js hosted on a CDN, isn't hosting plugins locally defeating the purpose? 

I think the issues you raise are interesting but I would file a separate issue for them and merge this PR. What do you think?

a4z:
I use the CDN especially during creation of presentation but also for talks company intern since I switch computers very often, and this makes things  easy. so far I never encountered any problem, but for important talks I have everything as a bundle.
yes, copying the plugin to a user defined folder. like custom plugins, and load it from there work with revealjs via CDN. Just not load it twice, with the default path, and a custom one.
That's why I had to change the default template, and when I was there I also removed the plugins I do not need because things load slightly faster if this is omitted.

I am thinking about striping out the whole plugin configuration section into a file, and load it, since I have always a config file. And most of the default loaded plugins are not used by me so I see no point in loading them, while I add other.
I fixed this 10 minute before a talk in the company I am working, and did not have time to thing about since then.
And currently I am on vacation, so I think I will not have something this week,

but the first thought is, why not strip out the whole section if there is something custom give, otherwise use the default ....
it should theoretical be possible to read an other file if no file is given, so ... 
will try it after vacation

a4z:
Thanks for your effort and improvements @obilodeau !!

PR #203, opened 2018-08-31
add what I need to do to run tests

a4z:


obilodeau:
Great catch! Sorry, it wasn't accurate.

PR #207, opened 2018-10-16
resolves #16 allow background color of slide to be set using CSS

### LINKED ISSUE #16: Allow slide background color to be set using a role
The slide background color can only be set by hard-coding the value in the `data-background-color` attribute. This requirement has the following drawbacks:
- it clutters up the template
- it diminishes the semantic quality of the source document
- it ties the styling to a single theme

Better would be to allow the background color to be set using a role. The role introduces a layer of indirection between the content and the presentation and it can be used to change the color across all corresponding slides.

Unfortunately, reveal.js doesn't support setting the background color of a slide via CSS. The reason is because reveal.js reads the background color directly from the `data-background-color` attribute on the section and applies it to a floating canvas layer. If you attempt to set the background color of a slide using CSS, you'll notice that it only colors the content portion of the slide.

Since we have complete control over the generated HTML, we can patch in a solution that allows the background color to be controlled using a role. I've come up with two possible solutions.
### Solution 1: Read the color value from an attribute

If we agree on preset section role names, we could map the role to an AsciiDoc attribute and assign that value to the `data-background-color` attribute on the section when the HTML is generated. Here's an example:

```
= My Presentation
Presenter Name
:topic-background-color: red

[.topic]
== Topic Slide

== Content Slide

content
```

The resulting section element for the "Topic Slide" would be something like:

```
<section class="topic" data-background-color="red"><h2>Topic Slide</h2></section>
```

We could follow a fixed pattern so that we don't have to decide on preset class names (something like `section-*`).
### Solution 2: Use JavaScript to assign the `data-background-color` dynamically

We could inject JavaScript that runs after the presentation loads to read the effective background of each slide (from calculated styles) and apply those background-color values to the `data-background-color` attribute while removing the background-color from the section. This is the cleanest approach, but will require some research.

---

I'm happy with Solution 1 until we can get something more formal in place. Perhaps we can implement as two steps since Solution 1 honors the same contract in the AsciiDoc content as Solution 2...it just requires a shim (the document-level color attribute).

mojavelinux:


obilodeau:
Awesome! I'll work on docs, passing tests and merging this before our next release.

obilodeau:
Merge commit polluted this PR, see #220 for a clean one. Closing.

PR #213, opened 2018-10-29
Simpler image right alignment

### LINKED ISSUE #197: role=right on image block's behavior could be improved
Odd corner case I faced today. I usually do floating images this way:

```asciidoc
== Slide Title

* Some
image:ego-market_paper_cover.png[role=right,width=200px]
* Points
* I'm trying
* To make
```

But in this case it is ugly:

imagespan-role-right

So trying various things, I think applying a role=right on an image block *above* the main slide content should do what I want (and be expected behavior for people familiar with html/css).

```asciidoc
== Slide Title

image::ego-market_paper_cover.png[role=right,width=200px]

* Some
* Points
* I'm trying
* To make
```

This gives:

imageblock-role-right-now

But it's because there is no CSS instruction for role=right. I manually added one:

imageblock-role-right

This is I think what we want. Thoughts?

obilodeau:
A "right" role can be applied on all divs under revealjs CSS class namespace. If you think this is a bad idea, say so. This is our first attempt at overriding CSS from reveal.js.

Initial attempt is promising, we could backport a lot of AsciiDoc's features like tables and more. Lots of issues already open for these.

That said, I would have preferred a pure `.css` file include (so that syntax highlighting and IDE would have played along better) but I couldn't make it work.

PR to fix #197.

Note to reviewers: there is more test code in there than actual code. I'll merge it relatively soonish so I can get my Hacktoberfest t-shirt :wink:


mojavelinux:
:+1: 

When considering how to implement styles, don't rule out flexbox for alignment. It's supported perfectly across browsers and it it's extremely powerful. In particular, it's very good at self-alignment of items.

PR #215, opened 2018-10-30
Clarified in docs and examples what's possible with floating and roles

obilodeau:
New section about Asciidoctor-reveal.js specific roles, took the opportunity to introduce how roles are written in AsciiDoc and linked to upstream docs. Also took care of mentioning that image macro have a slightly different behavior.

Added examples (and tests) too.

Related to the just-added #213

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -495,7 +495,7 @@ Or
 See https://asciidoctor.org/docs/user-manual/#role[Asciidoctor's documentation] for more details.
 
 .Image specific note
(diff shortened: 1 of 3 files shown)


obilodeau:
Just want to note that only the upstream tests have failed and this is due to Asciidoctor introducing drastic build changes today. These build problems are temporary.

obilodeau:
Since these touches only docs plus examples and not code, I'm going to merge this before our usual 24 hour wait window.

PR #216, opened 2018-10-30
Attempt at making Travis happy about Asciidoctor 2.0

obilodeau:


obilodeau:
Some issues are upstream like in asciidoctor-doctest. I'm sure they'll get handled in a timely fashion. Not that we are in a rush, 2.0 dev started today so it shouldn't appear overnight. Yay CI-testing with upstream in mind :smile:

mojavelinux:
For compatibility with 1.5.8, you should add a dependency on the concurrent-ruby gem. You can still depend on thread_safe too. As long as both gems are present, it will work on both 1.5.7 and 1.5.8 without any warnings.

For compatibility with 2.0.0, you'd drop (exclude) the following rvms from the upstream build matrix:

* 2.2
* 2.1

Aside from that, I expect everything to work the same.

obilodeau:
If you look at Travis, you will see that the problem is asciidoctor-doctest
using `~> 1.5.0` (like we were doing). Sent them a PR.

Curious, why is the dep on thread_safe or concurrent isn't done upstream in
your gemfile? Why do we have to specify it downstream?

Le mar. 30 oct. 2018 2 h 28 a.m., Dan Allen <notifications@github.com> a
écrit :

mojavelinux:
Ah,  yes, that makes sense. 2.0.0 will not introduce any API changes, so it's safe to upgrade just as you would a 1.5.x release. We're just reestablishing the major version so that we can start following semantic versioning from that point forward.

Still, you'll want to drop 2.1 and 2.2 as they are no longer officially supported by the Ruby language project.

> why is the dep on thread_safe or concurrent isn't done upstream in your gemfile? Why do we have to specify it downstream?

Asciidoctor core has a policy of having 0 dependencies. On top of that, it's only a recommended dependency, not a required one.

And, btw, it's now concurrent-ruby, not thread_safe. Though, for it to work without warning on both 1.5.7 and 1.5.8, you'll need both on the load path.

Gemfile.upstream
```diff
@@ -3,3 +3,5 @@
 source 'https://rubygems.org'
 gemspec
 gem 'asciidoctor', :git => 'https://github.com/asciidoctor/asciidoctor', :branch => 'master'
+# TODO: remove once asciidoctor-doctest > v2.0.0-beta4 is shipped
+gem 'asciidoctor-doctest', :git => 'https://github.com/asciidoctor/asciidoctor-doctest', :branch => 'master'
```
mojavelinux: Good interim solution!

asciidoctor-revealjs.gemspec
```diff
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.summary = 'Converts AsciiDoc to HTML for a Reveal.js presentation'
   s.description = 'Reveal.JS back-end for Asciidoctor and Asciidoctor.js. Write slides in AsciiDoc!'
   s.license = 'MIT'
-  s.required_ruby_version = '>= 2.3.8'
+  s.required_ruby_version = '>= 2.3.0'
```
mojavelinux: My preference is to not set the required Ruby version, just document it in the README. The reason is, there's no way to override the required Ruby version at install time, and this can break builds. If you disagree, that's fine. I'm just sharing what I do.

obilodeau: Fixed with 4fce207. Followed your advice.

PR #219, opened 2018-11-02
Js embed modern frameworks

ggrossetie:
I can't push on your branch so I've created a new pull request based on your work with two fixes.

ping @obilodeau


test/smoke.js
```diff
@@ -1,19 +1,25 @@
...
 
 * Foo
 * Bar
-* World`;
+* World`
+
+const result = asciidoctor.convert(content, options)
 
-const result = asciidoctor.convert(content, options);
+expect(asciidoctorRevealjs.getVersion()).to.be('1.1.4-dev')
```
Mogztter: We should probably use a variable, otherwise we will have to update this value when we release a new version :disappointed:

src/asciidoctor-revealjs.tmpl.js
```diff
@@ -0,0 +1,46 @@
...
+   */
+  function getVersion () {
+    return resolveModule().$$const.VERSION.toString()
+  }
+
+  /**
+   * Registers the reveal.js converter.
+   */
+  function register () {
+    return resolveModule()
```
Mogztter: The function resolves the module (ie. initialize the module only once).

lib/asciidoctor-revealjs.rb
```diff
@@ -1,5 +1,6 @@
 if RUBY_ENGINE == 'opal'
   require 'asciidoctor-revealjs/converter'
+  require 'asciidoctor-revealjs/version'
```
Mogztter: `version` is required so we can get the version of the converter from the API.

```js
const revealjsConverter = require('asciidoctor-reveal.js')
console.log(revealjsConverter.getVersion()) // 1.1.4-dev
```

src/asciidoctor-revealjs.tmpl.js
obilodeau: I'm not the one you should ask that unfortunately :(

test/smoke.js
obilodeau: Sourced it out of `package.json` in 289e390. I think it should work in all cases.

src/asciidoctor-revealjs.tmpl.js
Mogztter: @obilodeau I think it's fine because it can be useful 😉

obilodeau (approved):

test/js-babel/README.adoc
```diff
@@ -0,0 +1,29 @@
...
+2. Install devel npm package with:
++
+ $ npm i --save ../../asciidoctor-reveal.js-1.1.4-dev.tgz
+
+Test the project:
+
+ $ npm test
+
+[IMPORTANT]
+====
```
obilodeau: I forgot to keep js-babel in sync with new changes. Will need to update it

obilodeau (changes requested):
New Todo item

test/js-babel/README.adoc
obilodeau: Fixed with 52c4e3f

obilodeau (approved):

test/js-babel/main.js
```diff
@@ -0,0 +1,21 @@
+import Asciidoctor from 'asciidoctor.js'
+const asciidoctor = Asciidoctor()
+
+// BEFORE SUPPORT
+// Babel will move the import statement at the top of the file.
+// As a consequence, the reveal.js converter will be loaded before the initialization of Asciidoctor.js.
+//import 'asciidoctor-reveal.js'
+
+// NOW SUPPORTED
+import AsciidoctorRevealJs from 'asciidoctor-reveal.js'
```
Mogztter: Just a nitpick but I think the project name is lowercase `reveal.js`, so the variable should be `AsciidoctorRevealjs` ?

ggrossetie (commented):
I can't approve (my own pull request) but it looks good

test/js-babel/main.js
obilodeau: sure, why not

PR #231, opened 2018-12-08
drop sudo: false from Travis CI config

mojavelinux:
remove sudo: false (container-based infrastructure) as recommended by Travis

See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


obilodeau:
Thanks for taking the time to do the PR!

PR #240, opened 2019-02-06
Javascript/Node: Sync with Asciidoctor.js 1.5.9 (fixes #227)

### LINKED ISSUE #227: Sync to Asciidoctor.js' 1.5.9 (and matching Opal)
We post-poned #187. It means we specify asciidoctor.js' dependency still. A major release means we should sync with the latest code.

Opal is only required as a devDep, make sure it's not pulled-in in ruby environments if we specify a special version in the Gemfile.

obilodeau:


obilodeau:
Just a note to say that the build failed because of the _upstream_ Gemfile pulling in upcoming Asciidoctor 2.0. For the current asciidoctor release, the build passed.

mojavelinux:
I'll have a look. My goal was not to impact existing converters, but there's something I'm missing.

PR #243, opened 2019-02-28
Travis: Added Ruby 2.6 as a target

obilodeau:
Adding ruby 2.6 and once our 2.0 is released, we will drop older rubies following upstream Asciidoctor's plan.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ git:
 language: ruby
 node_js: lts/*
 rvm:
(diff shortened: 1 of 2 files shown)


obilodeau:
Note: Build failures are for Asciidoctor master (and are not really failures I think)

obilodeau:
Force pushed because I made build fixes on master and want to merge with the tests passing.

PR #247, opened 2019-03-02
Dropping Ruby 2.1 and 2.2 (End of Life)

obilodeau:
Will merge only once I know I'm doing a feature release. Don't want to drop support on a bugfix release.


mojavelinux:
You probably want to go ahead and remove it from the build through (at least the upstream build) since Asciidoctor no longer supports Ruby < 2.3 as of Asciidoctor 2.0.0.

PR #255, opened 2019-04-01
resolves #253 register for reveal.js

### LINKED ISSUE #253: Register this converter for "reveal.js"
Since the project name is "reveal.js" I think this converter should be registered for "revealjs" (for backward compatibility) and "reveal.js".

ggrossetie:


obilodeau:
Why do we want this?

obilodeau:
Ah I found #253, my bad. When it's in the PR's title the link is not explicitly made by GitHub...

obilodeau:
TODO (note to self): update examples and docs to reflect new name

Note for changelog: This is not a breaking change since the old name will still work.

obilodeau (changes requested):
Please update README code snippets with new name `reveal.js` and it'll be good.

Mogztter:
Good thing that you asked me to update the documentation!

I've noticed that if you define `-b reveal.js` then the "backend" attribute generated by the processor is invalid (ie. an attribute must only contain word characters and hyphens, so `backend-reveal.js` is invalid).

While this change is fully backward compatible, because the converter is still registered for `revealjs`, it does not add any value and can be confusing if the user wants to use `-b reveal.js`.

So I think we should hold off for now (until we found a good solution): https://github.com/asciidoctor/asciidoctor-reveal.js/issues/253#issuecomment-559408020

Mogztter:
Closing for now as it will cause too much disruption.

PR #267, opened 2019-06-26
fix: only use lowercase asciidoc attributes

bentolor:
Asciidoctor currently does only support/expect attributes to be
in lower case. Unfortunately the code expected attributes like
ealjs_defaultTiming which were never actually resolved and
always provided the default value

### DIFF
--- a/test/doctest/revealjs-features.html
+++ b/test/doctest/revealjs-features.html
@@ -80,7 +80,7 @@
     fragments: true,
     // Flags whether to include the current fragment in the URL,
     // so that reloading brings you to the same fragment position
-    fragmentInURL: false,
+    fragmentInURL: true,
     // Flags if the presentation is running in an embedded mode,
     // i.e. contained within a limited portion of the screen
     embedded: false,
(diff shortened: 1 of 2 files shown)


bentolor:
The failure is out of my control

> The requested url does not exist(22): 'https://repo1.maven.org/maven2/org/jruby/jruby-dist/9.2.6.0/jruby-dist-9.2.6.0-bin.tar.gz?rvm=1.29.8'

ebousse:
hi! any progress regarding this PR by any chance? asciidoctor-reveal.js is a little difficult to use without this fix, unfortunately.

thanks and good luck :)

test/doctest/revealjs-features.html
```diff
@@ -80,7 +80,7 @@ <h2>Second slide</h2>
     fragments: true,
     // Flags whether to include the current fragment in the URL,
     // so that reloading brings you to the same fragment position
-    fragmentInURL: false,
+    fragmentInURL: true,
```
obilodeau: Why is this change here? Can you roll it back or provide reasoning why it should be in?

bentolor: @obilodeau It's dead-simple: Because it's wrong. ;-)

Due to my PR the [option in the source actually takes effect](https://github.com/asciidoctor/asciidoctor-reveal.js/blame/master/examples/revealjs-features.adoc#L7): And there it says `true`, so the expected result shall be also `true`

bentolor:
Rebased to retrigger build with false-positive build failure: Now everything green

PR #268, opened 2019-06-27
feat: upgrade bundled font-awesome from 4.3.0 → 5.8.2

bentolor:
I was missing quite a bunch of new icons which appeared
in https://fontawesome.com/icons?d=gallery&m=free but
did not work.

This udpates font awesome to the latest CDN-available
version

### DIFF
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -25,7 +25,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     include asciidoctor_revealjs.css.slim
     - if attr? :icons, 'font'
       - if attr? 'iconfont-remote'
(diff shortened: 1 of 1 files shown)


Mogztter:
As mention by Dan, I believe that FontAwesome 5 is not 100% backward compatible with FontAwesome 4 (ie. some icons names changed).
So we should probably add a note on the CHANGELOG.

mojavelinux:
That's correct. For the record, Asciidoctor PDF is already on FontAwesome 5, so I'm okay with other packages in the ecosystem using it. We just have to be very careful with core since that's where the highest expectation is for things not to break. And if other packages move, it will make the transition in core easier.

obilodeau:
I'm surprised this didn't break a bunch of tests. I see this is testing for the `iconfont-remote` attribute. I don't think we have test coverage on that. I'm going to need to dig a little bit deeper into that.

PR #277, opened 2019-11-25
feat: Add PDF export options

gquintana:
Add options described in https://github.com/hakimel/reveal.js/#pdf-export

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -136,7 +136,7 @@ First you must install and configure {uri-nodejs-download}[Node.js] on your mach
 
 Using npm:
 
-Create a directory to place slides. Initialize the directory to store npm packages within that directory. 
+Create a directory to place slides. Initialize the directory to store npm packages within that directory.
 
  $ echo {} > package.json               # eliminates warnings, use `npm init` if you prefer
  $ npm i --save asciidoctor-reveal.js
@@ -877,6 +877,16 @@ a|Number of pixels to move the parallax background per slide
 |<a valid CSS display mode>
 |The display mode that will be used to show slides.
 Defaults to *block*
+
+|:revealjs_pdfSeparateFragments:
+|*true*, false
+|In PDF export, put each fragment on a separate page.
+Defaults to *true*
+
+|:revealjs_pdfMaxPagesPerSlide:
+|<integer>
+|In PDF export, when a slide does not fit on a single page, maximum number of pages.
+Defaults to *1*
 |===
 
 If you want to build a custom theme or customize an existing one you should
--- a/README.adoc
+++ b/README.adoc
@@ -893,12 +893,12 @@ Defaults to *block*
 |<percentage value>
 | Factor of the display size that should remain empty around the content. Defaults to *0.1*
 
-|:revealjs_pdfSeparateFragments:
+|:revealjs_pdfseparatefragments:
 |*true*, false
 |In PDF export, put each fragment on a separate page.
 Defaults to *true*
 
-|:revealjs_pdfMaxPagesPerSlide:
+|:revealjs_pdfmaxpagesperslide:
 |<integer>
 |In PDF export, when a slide does not fit on a single page, maximum number of pages.
 Defaults to *1*
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -229,6 +229,12 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
         minScale: #{attr 'revealjs_minscale', 0.2},
         maxScale: #{attr 'revealjs_maxscale', 1.5},
 
+        // PDF Export Options
+        // Put each fragment on a separate page
+        pdfSeparateFragments: #{to_boolean(attr 'revealjs_pdfSeparateFragments', true)},
+        // For slides that do not fit on a page, max number of pages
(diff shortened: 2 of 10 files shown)


README.adoc
```diff
@@ -877,6 +877,16 @@ a|Number of pixels to move the parallax background per slide
 |<a valid CSS display mode>
 |The display mode that will be used to show slides.
 Defaults to *block*
+
+|:revealjs_pdfSeparateFragments:
```
Mogztter: You need to use lowercase:

```suggestion
|:revealjs_pdfseparatefragments:
```


See: https://github.com/asciidoctor/asciidoctor-reveal.js/commit/134bf3d29b82db787db069e30290144463e4f033

README.adoc
```diff
@@ -877,6 +877,16 @@ a|Number of pixels to move the parallax background per slide
...
+|*true*, false
+|In PDF export, put each fragment on a separate page.
+Defaults to *true*
+
+|:revealjs_pdfMaxPagesPerSlide:
```
Mogztter:
```suggestion
|:revealjs_pdfmaxpagesperslide:
```

templates/document.html.slim
```diff
@@ -229,6 +229,12 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
         minScale: #{attr 'revealjs_minscale', 0.2},
         maxScale: #{attr 'revealjs_maxscale', 1.5},
 
+        // PDF Export Options
+        // Put each fragment on a separate page
+        pdfSeparateFragments: #{to_boolean(attr 'revealjs_pdfSeparateFragments', true)},
```
Mogztter:
```suggestion
        pdfSeparateFragments: #{to_boolean(attr 'revealjs_pdfseparatefragments', true)},
```

templates/document.html.slim
```diff
@@ -229,6 +229,12 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
...
+        // For slides that do not fit on a page, max number of pages
+        pdfMaxPagesPerSlide: #{attr 'revealjs_pdfMaxPagesPerSlide', 1},
```
Mogztter:
```suggestion
        pdfMaxPagesPerSlide: #{attr 'revealjs_pdfmaxpagesperslide', 1},
```

Mogztter:
Tests are failing because the expected output changed: https://travis-ci.org/asciidoctor/asciidoctor-reveal.js/jobs/616847269?utm_medium=notification&utm_source=github_status

Ideally we should also add tests on this new feature to make sure that `pdfMaxPagesPerSlide` has the correct value when `revealjs_pdfmaxpagesperslide` is defined. Same thing applies for `pdfSeparateFragments`.

README.adoc
gquintana: There are already revealjs_parallaxBackgroundHorizontal, revealjs_showNotes, etc. most options are revealjs_camelCaseOption. The only exceptions are revealjs_minscale and revealjs_maxscale. It doesn't make sense to me lowercase the option.
Are you sure you want me to change this?

Mogztter: > There are already revealjs_parallaxBackgroundHorizontal, revealjs_showNotes, etc. most options are revealjs_camelCaseOption. The only exceptions are revealjs_minscale and revealjs_maxscale. It doesn't make sense to me lowercase the option.

It has changed recently, now all the attributes are lowercased (on master).

> Are you sure you want me to change this?

Yes.

> Although uppercase characters are permitted in an attribute entry (the place where an attribute is defined), the attribute name is converted to lowercase before being stored. The attribute name in an attribute reference is also converted to lowercase before the attribute is resolved. For example, URI, Uri and uRI are all treated as uri. (See issue #509 for a proposed change to this restriction). A best practice is to only use lowercase for letters in the name and avoid starting the name with a number.

https://asciidoctor.org/docs/user-manual/#attribute-restrictions

In short, you can define `:revealjs_camelCaseOption:` in your document but the attribute name will be converted to `revealjs_camelcaseoption`. So in the template we must use lowercase.
To avoid confusion, we also use lowercase in the documentation.

gquintana: @Mogztter On the branch master, it's still camel case, there's something I don't understand? Your link is pointing to asciidoctor manual, not asciidoctor-reveal.js.

Mogztter: Do you have a link where it's still in camel case? Here's what I get on the master branch: 

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/45a02329209d5a48a2ba6932102376189f35c222/templates/document.html.slim#L128-L244

> Your link is pointing to asciidoctor manual, not asciidoctor-reveal.js.

Yes but Asciidoctor Reveal.js is using Asciidoctor to resolve attributes so we must comply with the above rule.

gquintana: Here https://raw.githubusercontent.com/asciidoctor/asciidoctor-reveal.js/master/README.adoc

Mogztter: Oh I see... technically it works because `revealjs_camelCaseOption` will be resolved to `revealjs_camelcaseoption`.

@obilodeau Do you have an opinion? Should we also use lowercase in the documentation?

obilodeau: Yeah, at first I created a commit that fixed the case for all the doc but then I realized: why bother? It worked for people before and it will continue working. If attributes ever become case-sensitive then we will fix it and in the meantime it is more readable than without casing.

PR #278, opened 2019-11-28
Fix reveal.js capitalization and remove Asciidoctor in the package.json description

ggrossetie:


mojavelinux:
Doesn't the converter still work with Asciidoctor Ruby? Removing Asciidoctor seems to indicate it only works with Asciidoctor.js.

Mogztter:
I removed it from the `package.json`, Asciidoctor is still mentioned in the Gem description/summary. The node package only works with Asciidoctor.js and in my opinion it's confusing to say that the converter is for Asciidoctor and Asciidoctor.js when in fact you can't use the Node package with Asciidoctor Ruby.

If we talk about the "project" then we should not mention the implementation and say "A reveal.js converter for Asciidoctor".
In this case, we expect the user to understand that since it's a Node package it's going to work with the JavaScript implementation of Asciidoctor (implicit).

On the project page it makes sense to say that the converter works for both Asciidoctor and Asciidoctor.js (we could also add that it works with AsciidoctorJ) but on the packages (jar, gem, npm...) I think we should not talk about the other implementations.

mojavelinux:
That makes total sense. :man_facepalming:

obilodeau:
> I removed it from the `package.json`, Asciidoctor is still mentioned in the Gem description/summary. The node package only works with Asciidoctor.js and in my opinion it's confusing to say that the converter is for Asciidoctor and Asciidoctor.js when in fact you can't use the Node package with Asciidoctor Ruby.
> 
> If we talk about the "project" then we should not mention the implementation and say "A reveal.js converter for Asciidoctor".
> In this case, we expect the user to understand that since it's a Node package it's going to work with the JavaScript implementation of Asciidoctor (implicit).
> 
> On the project page it makes sense to say that the converter works for both Asciidoctor and Asciidoctor.js (we could also add that it works with AsciidoctorJ) but on the packages (jar, gem, npm...) I think we should not talk about the other implementations.

Very good points. I'll try to remember that nuance. Feel free to correct me if I mix things up in the future.

Issue #262, opened 2019-06-05
Listing some good template/example/top-notch presentations


danyill:
A few examples of top-notch open source asciidoctor-reveal.js presentations would help the casual user get started and provide a good base for hacking a presentation together. If people are like me they don't have time to do the infrastructure work because they started their presentation well behind schedule and just want something to start from which works and hopefully shows a good syntax range.

A couple of suggestions:

* https://github.com/melix/gradle-presentation-template
* https://gitlab.com/jorge-aguilera/presentation-template

I'd like to add a Gulp+Live Reload template as well, perhaps in some time (mine is not suitable for publication at present).

Could we gather a few and include them in the readme?

bentolor:
Not sure if top notch, but I'm really suprised that I sticked with asciidoctor-revealjs until the _very end_. At first my idea was to use it only for drafting and then to improve manually.

Nevertheless my presentation has yet _quite_ a few hacky/nasty parts and it also requires #267 #268 #269 to be merged before it will start to works.

All kind of reveal-js crazyness like backgroudn movies, fragments etc. included. Best watched in Firefox: Chrome sometimes fails with the vids. 

[So if you ever were interested into learning the news from Java 9 to 13 and up this presentation is for you ;-)](https://bentolor.github.io/java9to13/)

[asciidoctor-revealjs Source](https://github.com/bentolor/java9to13/tree/master/docs)

obilodeau:
Yes @danyill, it's a good idea. Feel free to start a README section in a PR or wait for me to get to it eventually.

> [So if you ever were interested into learning the news from Java 9 to 13 and up this presentation is for you ;-)](https://bentolor.github.io/java9to13/)

Oh my god, that's a very pretty presentation! Congrats! You inspire me to step up my game now :smile:. Videos in backgrounds for transitions or topic switches is such a good idea!

bentolor:
I think ideally visual guide/catalog (including screenshots of the result) would be helpful. What would be your stance about storage location for screenshots @obilodeau?

Or even better: how about introducing/showing of asciidoctor-reveal.js with a homepage of its output: _asciidoctor-reveal.js presenting „asciidoctor-reveal.js“_

obilodeau:
I like what was done in PR #268 

> Or even better: how about introducing/showing of asciidoctor-reveal.js with a homepage of its output: _asciidoctor-reveal.js presenting „asciidoctor-reveal.js“_

[...]

PR #284, opened 2019-12-04
Example showcase project, part of #262

danyill:
@bentolor I think your presentation is very worthy, would it be OK to add it to the showcase as per @obilodeau's suggestion? I like the idea of only a few words, some screenshots and some links.

I might hunt for a few other presentations on the weekend if this is about right.


bentolor:
Awesome! I feel honored. Feel free to include my presentation as a first starter. Looking forward to see a fee more light-weight examples for people as easy starter

PR #287, opened 2019-12-09
 resolves #261 use the new syntax highlighter API

### LINKED ISSUE #261: source highlighter rouge is not working with asciidoctorj 2.0.0
```
<sourceHighlighter>rouge</sourceHighlighter>
<attributes>
<rouge-css>style</rouge-css>
```                                  
I tried with asciidoctor-revealjs 2.0.0 and revealjs 3.5.0 as suggested in the asciidoctor maven examples.

It renders as 
```
<pre class="highlight listingblock"><code class="java language-java">
```

I expected rouge support to be working with revealjs.

ggrossetie:
This new API was introduced in Asciidoctor 2.0.0.
If the `syntax_highlighter` method is not available fallback to the "legacy" syntax highlighting.

When using the new syntax highlighting API, code blocks are encapsulated in a `<div class="listingblock">` and `<div class="content">`. I think it might be possible to remove them using the "transform" option on the `pre` and `code` elements to add the missing classes/attributes.

Not sure if the "noescape" data attribute is mandatory or not but when the syntax highlighter is "highlight.js", the code element will have the following attribute `data-noescape="true"`.

I think we should release a version 3.0.0 where we break compatibility with Asciidoctor < 2.0.0 because:
- we only test against the latest version of Asciidoctor
- the code is "duplicated" because we use two distinct approaches to do syntax highlighting
- it will reduce the maintenance


obilodeau:
> I think we should release a version 3.0.0 where we break compatibility with Asciidoctor < 2.0.0

Agreed. Please drop the legacy syntax highlighting support from this PR.

obilodeau:
Is it possible that `pygments` doesn't work under JRuby? If all the builds are going to fail after this is merged I would rather disable the `pygments` tests and document it. The asciidoctor docker image did something similar recently: https://github.com/asciidoctor/docker-asciidoctor/pull/63

Mogztter:
It's somehow working on Asciidoctor:

- https://github.com/asciidoctor/asciidoctor/blob/master/Gemfile
- https://github.com/asciidoctor/asciidoctor/blob/master/.travis.yml

I think the reason is that we are using Ubuntu xenial (with Python 3 I assume) and Asciidoctor is using Ubuntu trusty (with Python 2).
But I'm not sure why it's working with MRI... I might be wrong though...

If it's only failing in JRuby, I guess we could add a condition in the _Gemfile_:

```rb
group :development do
  gem 'pygments.rb' unless RUBY_ENGINE == 'jruby'
  gem 'rouge'
end
```

Or just remove the dependency on Pygments?

> At this point, we're recommending against the use of pygments as the source highlighter in Asciidoctor. Rouge is now supported by default in the major converters and doesn't rely on a system command or separate language runtime. So unless there is really good reason to continue supporting pygments in the Docker image, I'd make the argument that we should just drop it.

https://github.com/asciidoctor/docker-asciidoctor/pull/63#issuecomment-531657639

obilodeau:
I don't want to get rid of it since it does seem to work with MRI but I don't want it to make Travis fail since it's in bad shape (but that could be fixed by upstream eventually).

> If it's only failing in JRuby, I guess we could add a condition in the Gemfile
Unless I'm mistaken, only doing this is not going to make tests pass. That plus disabling the `pygments` doctest would probably make it happy.

If disabling doctest per engine is easy let's do that. Otherwise disabling all `pygments` test would bother me too much.

Mogztter:
@obilodeau I've added the `pygments.rb` dependency in the `if RUBY_ENGINE != 'jruby'` block.
Tests are green :tada:

obilodeau:
> Converting file examples/source-pygments.adoc... asciidoctor: WARNING: optional gem 'pygments.rb' is not available (reason: cannot load 'pygments'). Functionality disabled.
✔️;49m✔️

Fair enough :smile:

PR #290, opened 2019-12-10
Update Gem description and summary

ggrossetie:
To be consistent with the Node package and the repository description.


Mogztter:
Not related but IIRC you don't want to update the version in the gemspec right?

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/4bd85dffabcbcdf71ca4ba6d97ac8c998a3863cf/asciidoctor-revealjs.gemspec#L32

obilodeau:
Yes, update it, otherwise, syntax highlighting will stop working for users who don't update.

Mogztter:
@obilodeau Done! I've also updated the links in the "Minimum Requirements" section to point to Asciidoctor > 2.0.0.

Issue #252, opened 2019-03-31
Upgrade to Asciidoctor.js 2.0.0


Mogztter:
The final version is not yet available but I've published a second release candidate today. 
A few notes:

* The package was renamed to `asciidoctor`
* The package `asciidoctor` is now a meta-package that contains `@asciidoctor/core` and `@asciidoctor/cli`
* We probably should rename the package to `@asciidoctor/reveal.js-converter`
* We should make sure that the following is working:

```
$ npm i asciidoctor @asciidoctor/reveal.js-converter
$ asciidoctor -r @asciidoctor/reveal.js-converter -b reveal.js presentation.adoc
```

mojavelinux:
> `@asciidoctor/converter-reveal.js`

When I come back to this with fresh eyes, the order strikes me as awkward. As much as the programmer in me wants it to be inverted, I think we should use `@asciidoctor/reveal.js-converter` instead.

Mogztter:
If we want the following to be working, we will need to agree on an API.

```
$ asciidoctor -r @asciidoctor/reveal.js-converter -b reveal.js presentation.adoc
```

The reason is that the CLI will have to "automatically" register the converter.
We should also consider that we could register extensions:

```
$ asciidoctor -r @asciidoctor/reveal.js-converter -r @asciidoctor/plantuml-extension -b reveal.js presentation.adoc
```

In my opinion it's more intuitive than to require the user to create a JavaScript file to load and register the converters/extensions:

**init.js**
````js
require('@asciidoctor/reveal.js-converter').register()
````
```
$ asciidoctor -r ./init.js -b reveal.js presentation.adoc
```

obilodeau:
> > `@asciidoctor/converter-reveal.js`
> 
> When I come back to this with fresh eyes, the order strikes me as awkward. As much as the programmer in me wants it to be inverted, I think we should use `@asciidoctor/reveal.js-converter` instead.

@Mogztter's PR I just merged didn't contain the package name change. Since this release is going to be a major one due to the Node API change, I suggest we change the package name right now if it still makes sense to do so.

Should this be renamed `@asciidoctor/reveal.js-converter` or did our mind change about that?

[...]

PR #291, opened 2019-12-13
Publish the node package to the @asciidoctor organization

ggrossetie:
Package renamed from `asciidoctor-reveal.js` to `@asciidoctor/reveal.js`


obilodeau:
Unless I missed some conversation elsewhere. I would prefer if @mojavelinux explicitly agreed to that name? Following the discussion on #252.

mojavelinux:
:+1:

PR #294, opened 2019-12-18
Only include the relevant part when using doctest

ggrossetie:
I'm using the `:include:` directive to avoid copy-pasting the reveal.js inline script and the inline Asciidoctor default style.

I think we should have specific tests on the reveal.js inline script (to make sure that https://github.com/asciidoctor/asciidoctor-reveal.js/pull/267 won't happen again). I will do another pull request for that.


Mogztter:
Actually we do have tests (see: _revealjs-features.html_) so I'm not sure how #267 passed through the tests suite :thinking:

obilodeau:
Note: failing Travis tests are not related to our project

obilodeau:
Good to go

obilodeau:
For the record, tests were failing, they were fixed in another commit so I merged with master to see if tests were good. The PR doesn't look so good now...

PR #297, opened 2019-12-24
Alternate converter name: reduces surprises and backward compatible

obilodeau:
See #253

### DIFF
--- a/examples/multi-destination-content.adoc
+++ b/examples/multi-destination-content.adoc
@@ -50,6 +50,8 @@ Some overview
 ----
 
 ifdef::backend-revealjs[=== !]
+// TODO one day would be nice to allow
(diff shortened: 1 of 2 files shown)


examples/multi-destination-content.adoc
```diff
@@ -50,6 +50,8 @@ Some overview
 ----
 
 ifdef::backend-revealjs[=== !]
+// TODO one day would be nice to allow
+//ifdef::backend-reveal.js[=== !]
```
mojavelinux: This should already work. The `.` is legal in this context.

mojavelinux: But it will only be true if you activate the converter using `-b reveal.js` instead of `-b revealjs`. So it's either or. The nice part would be "and".

PR #299, opened 2019-12-27
Get the latest repository version of rvm

Mogztter:
Make sure that `jruby-9` will be resolved to an up-to-date version of jruby.


Mogztter:
`before_install` runs after the `rvm` matrix :disappointed:

Issue #296, opened 2019-12-23
Support reveal.js 3.8 and 3.9.1


obilodeau:
Note to self: There's probably not much to do here. But at least get the javascript requirements bumped and verify for new options.

obilodeau:
I started working on this today. There are a few new options and plugins support will need to be verified since the API changed upstream but it looks like it is backward compatible.

obilodeau:
After more work, the fact that upstream removed the `head.min.js` inclusion worries me. If we adopt reveal.js 3.8, the slides rendered by us will not be viewable by reveal.js 3.7.

The question becomes: is becoming incompatible with an older reveal.js release a break in API? For the javascript ecosystem, our `package.json` does specify the proper version although users are free to override. For the Ruby folks, if they followed the README they manage it by hand but with bundler our version should not bump for no reason.

If we consider stopping our support of 3.7 (and earlier) to be API breaking then we should do it during the 3.0 window. However, I don't expect many people to be forking their reveal.js so they should be reasonably able to upgrade. If so, it's not really an API break but more of a dependency update that is required. Then this would mean going smoothly by releasing 3.0.0 which works with 3.7 and then 3.1.0 which breaks 3.7 would be a better path forward for them.

My take: I think doing a 3.0.0 without reveal.js 3.8 (which works right now btw, just not the new features) and supporting it later for 3.1.0 would be reasonable but I would be interested in your thoughts @Mogztter, @mojavelinux, and others.

mojavelinux:
I'm for a major version bump when reveal.js is upgraded, no matter what version we're coming from. Let's not cut corners when it comes to semver.

Mogztter:
> If we adopt reveal.js 3.8, the slides rendered by us will not be viewable by reveal.js 3.7.

It sounds like a major version.

> I'm for a major version bump when reveal.js is upgraded, no matter what version we're coming from. Let's not cut corners when it comes to semver.

I agree 👍

obilodeau:
I agree with you guys. Then I think I'm going to post-pone reveal.js 3.8 support to an asciidoctor-reveal.js 4.0.0 release (which I can do soon).

[...]

PR #301, opened 2019-12-31
Reveal.js 3.8.0 to 3.9.2 support

obilodeau:
Draft commit for reveal.js 3.8.0 support.

This work is postponed to have a very stable 3.0.0 that will support reveal.js versions spanning from 3.0 to 3.7. See #296.

Needs work:
- [x] syntax highlight `<mark>` syntax [described here](https://github.com/hakimel/reveal.js/blob/master/README.md#code-syntax-highlighting) does not work
- [x] line number highlights [described here](https://github.com/hakimel/reveal.js/blob/master/README.md#line-numbers--highlights) do not work
- [x] default plugin call for highlightjs is not current but it doesn't seem to break anything
- [ ] ~I'm not sure per-slide iframe preloading options are supported~
- [x] Also, see commits messages and introduced comments for more to add to the task list
- [x] test `hash` feature w/ and w/o history and see if direct links still works


obilodeau:
* Rebased on top of 3.1.0
* Aligned with latest reveal.js `index.html` template for slide decks (dropped IE hacks, safari viewport stuff, etc.)
* Highlightjs changes (the way its called, default css to align with upstream)
* Dropped redundant print-pdf plugin call

Also, highlightjs `<mark>` does work if we put valid html angle brackets in the generated source file so we need to tell Asciidoctor to let them through. Result is ugly though but it seems to be intended (I tested with only reveal.js).

obilodeau:
Tonight's work:
* highlightjs `<mark>` works with `subs="none"`, adjusted the example accordingly
* highlightjs line highlight works, it kicks-in line highlighting which might be unexpected and there seems to be a CSS activated that breaks the output. Still need to work on this.

highlightjs-highlight-ugly

* added broken examples for other supported highlighters, given I have little time to spend on this project (and I'm spending a lot right now) I won't focus on them but they are documented as broken and people can step in

lib/asciidoctor-revealjs/highlightjs.rb
```diff
@@ -11,10 +11,20 @@ def initialize *args
...
+        end
+
         def format node, lang, opts
           super node, lang, (opts.merge transform: proc { |_, code|
             code['class'] = %(language-#{lang || 'none'} hljs)
             code['data-noescape'] = true
+            # Note for review: should the API be modified to give easier access to this?
+            if node.attributes.key?("highlight")
+              code['data-line-numbers'] = self._convert_linedef_to_revealjs(node.attributes["highlight"])
+            end
```
obilodeau: I don't think my access to the `highlight` attribute of the listing block is proper here. It should go through the syntax highlight API but I didn't find a way to do so. `opts[:highlight_lines]` is available in `highlight()` but not in `format()`.

@Mogztter, @mojavelinux thoughts?

examples/source-rouge.adoc
```diff
@@ -39,3 +40,20 @@ fn main() {
     println!("Hello stretched World!");
 }
 ----
+
+== Rouge Reatures
```
Mogztter:
```suggestion
== Rouge Features
```

examples/source-rouge.adoc
```diff
@@ -39,3 +40,20 @@ fn main() {
...
+
+// TODO docinfo not loaded
+// TODO linenums feature affects listing size
+
+This is broken
```
Mogztter: Actually, it's a limitation of Rouge: https://github.com/rouge-ruby/rouge/issues/264
I think we shouldn't bother too much.

examples/source-rouge.adoc
```diff
@@ -39,3 +40,20 @@ fn main() {
     println!("Hello stretched World!");
 }
 ----
+
+== Rouge Reatures
+
+// TODO docinfo not loaded
```
Mogztter: This sounds more concerning...

lib/asciidoctor-revealjs/highlightjs.rb
Mogztter: Sorry I can't help, I'm not familiar with this API so I don't know the reason behind this choice

Mogztter:
@obilodeau How can I help with this pull request? If you have specific tasks, please tell me. Hopefully I've fixed the docinfo issue in #324

examples/source-rouge.adoc
obilodeau: I'm not going to do it.

obilodeau:
Rebased on top of #324.

examples/source-rouge.adoc
obilodeau: docinfo is there now thanks to the rebase on top of #324. However, the class `hll` is missing from the styles inlined by Asciidoctor. Again, anything else than highlight.js is not getting my attention right now.

obilodeau:
Highlight.js is feature parity with 3.8.0. I'm going to deal with 3.8.0 to ~3.9.1~ 3.9.2 later.

Next-up: go through all configuration options and see if I'm missing anything

obilodeau:
* Added new configuration options
* Added `previewLinks` and `data-preview-link` support
* Rebased on top of #325
* Started upgrade notes and added a compatibility matrix to the README

obilodeau:
Pushed doctests

obilodeau:
Lazy loading is not supported at the moment and would require a lot of testing and changes. I have video-heavy presentations and never felt it was required. Also, it can be implemented later without causing a breaking change. Let's not implement it now.

obilodeau:
This is ready to go in. I have a big presentation coming up so I would like to release this relatively shortly (in about a week max).

@Mogztter if you want #320 to get in for 4.0.0, I would advise you to get it in shape quickly. I would prefer it if it does get in. Plus there is a lot of doctest around highlightjs so I feel confident.

@mojavelinux there's an API question that you might want to look at in this review

Lastly, @Mogztter since you asked how you can help, there's #326 that I would like to get done soon. I'm not very good at maintainable CSS so if you can give it a try I would appreciate it.

Mogztter:
Speaking of #326, I need to do some vertical alignment and as far as I know the only way to do it reliably is to wrap the content of a slide in a `<div>`. Is that something you would consider or should I override the "section" template and do it in my presentation?

obilodeau:
I had the feeling that it was achievable without doing so. Isn't it working
for the other example I provided in the issue?

Mogztter:
I don't want to pollute the pull request about the reveal.js upgrade, so here's my reply: https://github.com/asciidoctor/asciidoctor-reveal.js/issues/326#issuecomment-582480509

templates/document.html.slim
```diff
@@ -53,15 +55,15 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     - syntax_hl = self.syntax_highlighter
     - if syntax_hl && (syntax_hl.docinfo? :head)
       =syntax_hl.docinfo :head, self, cdn_base_url: cdn_base, linkcss: linkcss, self_closing_tag_slash: '/'
-    / If the query includes 'print-pdf', use the PDF print sheet
+
+    /! Printing and PDF exports
```
Mogztter: Do we really want an HTML comment here?

test/doctest/theme-custom.html
```diff
@@ -133,39 +165,37 @@ <h2>Actually, they are…​</h2>
     parallaxBackgroundVertical: null,
     // The display mode that will be used to show slides
     display: 'block',
-
+
```
Mogztter: can we remove the trailing spaces?

Mogztter:
> if you want #320 to get in for 4.0.0, I would advise you to get it in shape quickly. I would prefer it if it does get in. Plus there is a lot of doctest around highlightjs so I feel confident.

Alright, I will work on it tomorrow 😉

templates/document.html.slim
obilodeau: I'm exactly following reveal.js' presentation frame

test/doctest/theme-custom.html
obilodeau: Sure, `rake doctest:generate` creates them and I was too lazy to remove them. If `doctest` doesn't mind (and I my guess is it doesn't) then I'll remove them.

templates/document.html.slim
Mogztter: All good then 👍

test/doctest/theme-custom.html
obilodeau: Fixed in e5b8b64

obilodeau:
Fixed whitespace

lib/asciidoctor-revealjs/highlightjs.rb
mojavelinux: I didn't think about the fact that the option would be needed in format. We could take that up in core. In the meantime, you can call:

```
resolve_lines_to_highlight source, (attr 'highlight')
```

mojavelinux: don't do it this way:

```
highlight_lines.gsub("..", "-")
```

The logic is much more complex than that. See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/substitutors.rb#L968-L999

obilodeau: The problem is that reveal.js introduced a new legal character `|` to split highlights into steps sequences. It's a very cool feature that I intend to use in presentations. Documented here: https://github.com/hakimel/reveal.js#step-by-step-highlights

So I'm going to need to split on it, then send each section to `resolve_lines_to_highlight` method and rejoin everything after. I'm on it right now.

In the meantime, here how I adapted your code to make it work, in case you have any comment:

```ruby
if node.attributes.key?("highlight")
    code['data-line-numbers'] = node.resolve_lines_to_highlight(node.content, node.attributes["highlight"]).join(',')
```

obilodeau: Fixed in a839f53. This is probably the most idiomatic ruby I've written so far. Marking this as resolved but allowing you to further comment for another day before merging.

Issue #259, opened 2019-04-21
Create a statically linked ELF/PE/Mach-O binary


Mogztter:
I think it would be useful to add a binary in the npm package. If we do that, we would be able to execute the Reveal.js converter using `npx`:

```
$ npx -p @asciidoctor/reveal.js-converter asciidoctor-revealjs presentation.adoc
```

If we decide to publish a package with the same name as the command, then the command would be:

```
$ npx asciidoctor-revealjs presentation.adoc
```

Of course it would also be possible to install the package with `npm`:

```
$ echo {} > package.json        # eliminates warnings, use `npm init` if you prefer
$ npm i @asciidoctor/reveal.js-converter
$ ./node_modules/.bin/asciidoctor-revealjs presentation.adoc
```

akosma:
I'm currently using this script (written in TypeScript) to generate the HTML slides (executed using `ts-node`):

```typescript
// Drives the creation of the Reveal.js presentation using
// the asciidoctor-reveal.js JavaScript library.
import * as asciidoctorLib from 'asciidoctor.js';
import * as asciidoctorRevealjs from 'asciidoctor-reveal.js';
import * as fs from 'fs';
import * as path from 'path';

// Check that there is a parameter
if (process.argv.length < 3) {
    console.error('build.ts: Missing "filename" parameter. Exiting');
    process.exit(1);
}

// Check that the parameter points to a file that exists
const filename = process.argv[2];
const filepath : string = path.join(__dirname, '..', filename);
if (!fs.existsSync(filepath)) {
    console.error('parse_files.ts: The path "%s" is invalid. Exiting.', filepath);
    process.exit(1);
}

// Drive the transformation from Asciidoc to HTML
const asciidoctor = asciidoctorLib();
asciidoctorRevealjs.register();
const options = {safe: 'safe', backend: 'revealjs'};
asciidoctor.convertFile(filename, options);
```

obilodeau:
I think this goal was achived through the merger of #254. The current `master` branch allows the use of a binary to build slides:

```
$ $(npm bin)/asciidoctor -r asciidoctor-reveal.js -b revealjs presentation.adoc
```

There's even a way to do so from PowerShell.

Can I close this?

obilodeau:
> Can I close this?

After discussing with @Mogztter, I understand now what this aims to do. Updated the title of the ticket accordingly. Still aiming to achieve that shortly.

mojavelinux:
[...]

PR #308, opened 2020-01-14
Javascript CLI and binary packages

obilodeau:
Making progress but doesn't work right now.

Created the PR to get assistance from @Mogztter. 

Related to #259.


obilodeau:
I know there's a ton of crap in there. I'm trying to get it to work and then I'll cleanup.

Right now with `--verbose` I get:

```
$ npx asciidoctor-revealjs -v presentation.adoc 
require undefined
backend html5
doctype undefined
standalone true
section-numbers false
failure-level FATAL
quiet false
verbose true
timings false
trace false
base-dir undefined
destination-dir undefined
verbose-mode 2
attributes 
options {"backend":"html5","safe":"unsafe","standalone":true,"failure_level":4,"verbose":2,"timings":false,"trace":false,"mkdirs":true,"attributes":[]}
converting file presentation.adoc
```

backend is wrong, file is right but doesn't get converted

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,130 @@
...
+'use strict'
+
+process.title = 'asciidoctor-revealjs'
+
+const pkg = require('../package.json')
+const { Options, Invoker, processor } = require('@asciidoctor/cli')
+const asciidoctor = require('@asciidoctor/core')();
+const asciidoctorRevealjs = require('@asciidoctor/reveal.js')
+const path = require('path')
+const EventEmitter = require('events')
```
Mogztter: You don't need this part (it's an implementation detail of a Mutex). In this case, we don't need a Mutex.

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+
+
+//const { PdfOptions, PdfInvoker } = require('../lib/cli.js')
+
+;(async () => {
+//  const options = new RevealJsOptions().parse(process.argv)
+  //return new RevealJsInvoker(options).invoke()
+  const options = new Options().parse(process.argv)
```
Mogztter: You can pass default values to the `Options` constructor.
Here you need to pass the backend:

```suggestion
  const options = new Options({backend: 'revealjs'}).parse(process.argv)
```

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,130 @@
...
+      this.showHelp()
+      return { exit: true }
+    }
+  }
+}
+
+
+//const { PdfOptions, PdfInvoker } = require('../lib/cli.js')
+
+;(async () => {
```
Mogztter: I don't think you need this `async` block, as far as I remember, the code is not async.

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,130 @@
...
+  }
+})
+
+class RevealJsInvoker extends Invoker {
+  async invoke () {
+    const processArgs = this.options.argv.slice(2)
+    const { args } = this.options
+    const { verbose, version, files } = args
+
+    if (version || (verbose && processArgs.length === 1)) {
```
Mogztter: I don't think you need to override the `invoke` function.

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,130 @@
...
+      process.stdout.write(`Asciidoctor-Reveal.js ${pkg.version} using `)
+      Invoker.printVersion()
+      return { exit: true }
+    }
+
+    // Register the reveal.js converter
+    asciidoctorRevealjs.register()
```
Mogztter: The converter can be registered once at the top of the script.

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,130 @@
...
+})
+
+class RevealJsInvoker extends Invoker {
+  async invoke () {
+    const processArgs = this.options.argv.slice(2)
+    const { args } = this.options
+    const { verbose, version, files } = args
+
+    if (version || (verbose && processArgs.length === 1)) {
+      process.stdout.write(`Asciidoctor-Reveal.js ${pkg.version} using `)
```
Mogztter: I think I need to define `printVersion` as an instance function (instead of a static function).

```js
class RevealJsInvoker extends Invoker {

  printVersion () {
    process.stdout.write(`Asciidoctor-Reveal.js ${pkg.version} using `)
    super.printVersion()
  }
}

File: npm/asciidoctor-revealjs [hunk 7]
```diff
@@ -0,0 +1,130 @@
...
+.then((result) => {
+  if (result.exit) {
+    process.exit(0)
+  }
+})
+.catch((error) => {
+  console.log('error', error)
+  process.exit(1)
+})
+
```
Mogztter: Could you try the following?

```js
#!/usr/bin/env node

'use strict'

process.title = 'asciidoctor-revealjs'

const { Options, Invoker } = require('@asciidoctor/cli')
const asciidoctorRevealjs = require('@asciidoctor/reveal.js')
asciidoctorRevealjs.register()

const options = new Options({backend: 'revealjs'}).parse(process.argv)
new Invoker(options).invoke()
```

I think it should work, then we can address the issue with `printVersion`

npm/asciidoctor-revealjs
obilodeau: I can confirm. I added the code but it won't work until you call this dynamically: https://github.com/asciidoctor/asciidoctor-cli.js/blob/master/lib/invoker.js#L20.

Do you consider this a big change for the `@asciidoctor/cli` or are you willing to do it in a point or feature release?

obilodeau:
@Mogztter you should read #259 for struggles with `pkg`. I made the comments at the wrong place.

Last thing I just realized: I don't like the fact that we need both `npm/asciidoctor-revealjs` and `packages/asciidoctor-revealjs/bin/asciidoctor-revealjs` that are identical because the former is used as the CLI with npm/npx and the latter is used to build binaries (different dependencies that we don't want to force on users for the former). A solution for that would be awesome!

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+
+'use strict'
+
+process.title = 'asciidoctor-revealjs'
+
+const { Options, Invoker } = require('@asciidoctor/cli')
+const asciidoctorRevealjs = require('@asciidoctor/reveal.js')
```
Mogztter: You should use a relative path:

```suggestion
const asciidoctorRevealjs = require('../dist/main.js')
```

packages/asciidoctor-revealjs/package.json
```diff
@@ -0,0 +1,50 @@
+{
```
Mogztter: You should probably add `"private": true` and keep it to a minimum with only `bin`, `dependencies` and `devDependencies`.
You don't need all the project information since it won't be published.

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+
+'use strict'
+
+process.title = 'asciidoctor-revealjs'
+
+const { Options, Invoker } = require('@asciidoctor/cli')
```
Mogztter: Is it working? I'm surprise because you are using `require('@asciidoctor/cli')` but `@asciidoctor/cli` is not a direct `dependencies` in the root `package.json`?

packages/asciidoctor-revealjs/index.js
```diff
@@ -0,0 +1 @@
+module.exports = require('@asciidoctor/reveal.js')
```
Mogztter: I don't think you need this file.

packages/asciidoctor-revealjs/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "name": "@asciidoctor/reveal.js",
+  "version": "3.0.1-dev",
+  "description": "A reveal.js converter for Asciidoctor.js. Write your slides in AsciiDoc!",
+  "main": "index.js",
```
Mogztter: You can remove the `main` field. This project will only produce a binary.

packages/asciidoctor-revealjs/bin/asciidoctor-revealjs
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env node
+
+'use strict'
+
+process.title = 'asciidoctor-revealjs'
+
+const { Options, Invoker } = require('@asciidoctor/cli')
+const asciidoctorRevealjs = require('@asciidoctor/reveal.js')
```
Mogztter: I think we want to use the current project using a relative path.

Mogztter:
> Last thing I just realized: I don't like the fact that we need both npm/asciidoctor-revealjs and packages/asciidoctor-revealjs/bin/asciidoctor-revealjs that are identical because the former is used as the CLI with npm/npx and the latter is used to build binaries (different dependencies that we don't want to force on users for the former). A solution for that would be awesome!

On further consideration, I'm not sure we need to create an additional project to build the binary.
If we use a relative path to resolve the `@asciidoctor/reveal.js` dependency then we only need to install `@asciidoctor/core`.

Since we don't want to add `@asciidoctor/core` as a direct dependency, I guess we could install it just before we need to run `pkg` right?

**package-bin.sh**
```sh
npm i @asciidoctor/core
./node_modules/.bin/pkg -t "node12-linux-x64,node12-macos-x64,node12-win-x64" package.json --out-path dist/
```

Mogztter:
I've created a pull request: https://github.com/obilodeau/asciidoctor-reveal.js/pull/1 (to your branch)

npm/asciidoctor-revealjs
Mogztter: > Do you consider this a big change for the @asciidoctor/cli or are you willing to do it in a point or feature release?

I think I can ensure backward compatibility but I might need to use another name for the function.

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,21 @@
...
+const asciidoctorRevealjs = require('@asciidoctor/reveal.js')
+asciidoctorRevealjs.register()
+
+const options = new Options({backend: 'revealjs'}).parse(process.argv)
+
+class RevealJsInvoker extends Invoker {
+  printVersion () {
```
Mogztter: @obilodeau You can use `@asciidoctor/cli` version 3.1.0 and override the `version` function:

```js
version () {
  const pkg = require('../package.json')
  return `Asciidoctor reveal.js ${pkg.version} using ${super.version()}`
}
```

packages/asciidoctor-revealjs/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "Olivier Bilodeau (https://github.com/obilodeau)",
+    "Guillaume Grossetie (https://github.com/mogztter)"
+  ],
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/asciidoctor/asciidoctor-reveal.js/issues"
+  },
+  "homepage": "https://github.com/asciidoctor/asciidoctor-reveal.js",
+  "dependencies": {
+    "@asciidoctor/cli": "^3.0.2",
```
Mogztter:
```suggestion
    "@asciidoctor/cli": "^3.1.0",
```

package.json
```diff
@@ -39,11 +45,13 @@
   },
   "homepage": "https://github.com/asciidoctor/asciidoctor-reveal.js",
   "dependencies": {
+    "@asciidoctor/cli": "^3.1.0",
```
obilodeau: Isn't this a problem in the overall objective of allowing users to run their own asciidoctor.js version? Globally installed asciidoctor.js will not conflict or cause trouble?

README.adoc
```diff
@@ -148,14 +147,14 @@ You can now install the dependencies:
 
 === Convert AsciiDoc into slides
 
-Once the dependencies are installed, verify that the `asciidoctor` command is available.
+Once the dependencies are installed, verify that the `asciidoctor-revealjs` command is available.
 On Linux and macOS, open a terminal and type:
 
- $ $(npm bin)/asciidoctor --version
+ $ npx asciidoctor-revealjs --version
```
obilodeau: I saw that npx was core node now. Is it ok to use it? Is it in widely used distros now?

README.adoc
```diff
@@ -183,12 +182,12 @@ If you don't have an existing presentation, you can create a sample presentation
...
 To convert the sample presentation into slides, open a terminal and type:
 
- $ $(npm bin)/asciidoctor -r @asciidoctor/reveal.js -b revealjs presentation.adoc
+ $ npx asciidoctor-revealjs presentation.adoc
 
 On windows, open PowerShell and type:
 
- $ .\node_modules\.bin\asciidoctor.cmd -r @asciidoctor/reveal.js -b revealjs presentation.adoc
- 
+ $ .\node_modules\.bin\asciidoctor-revealjs.cmd presentation.adoc
```
obilodeau: Will this be good for this project or did you do something special for asciidoctor.js?

Mogztter: I'm not sure I quite understand your question... :thinking:

README.adoc
Mogztter: > I saw that npx was core node now. Is it ok to use it?

Yes :+1: 

> Is it in widely used distros now?

Not sure, it's included in `npm` so I guess it should be available when you install node/npm from your package manager.

package.json
Mogztter: The package `@asciidoctor/cli` does not have a direct dependency on `@asciidoctor/core`: https://github.com/asciidoctor/asciidoctor-cli.js/blob/7055e675d429ad6b9b0b57bf395d383a0cc6b6a3/package.json#L51-L53

README.adoc
obilodeau: The `.cmd` Windows file and the `\.bin\` directory are generated by npm? You didn't do anything specific to get that working?

Mogztter: > The .cmd Windows file and the \.bin\ directory are generated by npm?

Yes!

obilodeau:
The plan is to squash merge this in a few hours (later tonight EST). Last chance for comments! Unless someone requests to hold on it for a little while.

HACKING.adoc
```diff
@@ -274,6 +274,15 @@ See https://github.com/asciidoctor/asciidoctor-reveal.js/issues/187#issuecomment
 
 Asciidoctor.js versioning policy is https://asciidoctor-docs.netlify.com/asciidoctor.js/project/version-and-lifecycle-policies/[available here].
 
+=== Debugging
+
+To debug the javascript application, just add `--node-arg=--inspect-brk` to the npx command to run the application.
```
Mogztter:
```suggestion
To debug the JavaScript application, just add `--node-arg=--inspect-brk` to the npx command to run the application.
```

HACKING.adoc
```diff
@@ -274,6 +274,15 @@ See https://github.com/asciidoctor/asciidoctor-reveal.js/issues/187#issuecomment
...
+For example:
+
+  npx --node-arg=--inspect-brk asciidoctor-revealjs -v presentation.adoc
+
+The open the Chrome Dev Tools and click on the Node logo in the top left corner.
```
Mogztter:
```suggestion
Then open the Chrome Dev Tools and click on the Node logo in the top left corner.
```

HACKING.adoc
```diff
@@ -352,6 +361,8 @@ Then run:
  $ npm publish
 
 . Check that the new version is available on https://www.npmjs.com/package/@asciidoctor/reveal.js[npmjs.com]
+. Make binary releases
```
Mogztter: I'm not sure about this one, but there's one release of binaries? so:

```suggestion
. Make binaries release
```

README.adoc
```diff
@@ -216,6 +215,61 @@ To execute the script, open a terminal and type:
 
 You can open the `presentation.html` file in your browser and enjoy!
 
+
+== Standalone Executable
+
+Pre-built binary packages can be downloaded from our {uri-project-repo}/releases[GitHub release page].
+We provide them for Windows 64-bit, Linux 64-bit and MacOS 64-bit.
```
Mogztter:
```suggestion
We provide them for Windows 64-bit, Linux 64-bit and macOS 64-bit.
```

npm/asciidoctor-revealjs
```diff
@@ -0,0 +1,20 @@
...
+const { Options, Invoker } = require('@asciidoctor/cli')
+const asciidoctorRevealjs = require('../dist/main.js')
+asciidoctorRevealjs.register()
+
+const options = new Options({backend: 'revealjs'}).parse(process.argv)
+
+class RevealJsInvoker extends Invoker {
+  version () {
+    const pkg = require('../package.json')
+    return `Asciidoctor-Reveal.js ${pkg.version} using ${super.version()}`
```
Mogztter: I think the project name should be "Asciidoctor reveal.js" no?
```suggestion
    return `Asciidoctor reveal.js ${pkg.version} using ${super.version()}`
```

README.adoc
```diff
@@ -216,6 +215,61 @@ To execute the script, open a terminal and type:
...
+
+  $ git clone -b 3.7.0 --depth 1 https://github.com/hakimel/reveal.js.git
+
+=== Convert AsciiDoc into slides
+
+Open a terminal where the executable is and type:
+
+ $ ./asciidoctor-revealjs --version
+
+The command should report the Asciidoctor-Reveal.js and Asciidoctor CLI version in the terminal:
```
Mogztter:
```suggestion
The command should report the Asciidoctor reveal.js and Asciidoctor CLI version in the terminal:
```

README.adoc
```diff
@@ -216,6 +215,61 @@ To execute the script, open a terminal and type:
...
+
+[source,console]
+----
+Asciidoctor-Reveal.js 3.0.1 using Asciidoctor.js 2.0.3 (Asciidoctor 2.0.9) [https://asciidoctor.org]
```
Mogztter:
```suggestion
Asciidoctor reveal.js 3.0.1 using Asciidoctor.js 2.0.3 (Asciidoctor 2.0.9) [https://asciidoctor.org]
```

npm/asciidoctor-revealjs
obilodeau: I thought a lot about this before changing it from your code proposal. Here's the rundown:

* A project name rarely has spaces in them
* the space could make it non-obvious that it's the same thing, confusing users
* Most slide post-processors in the Asciidoctor ecosystem use a project name with dashes (bespoke)
* Reveal.js should be capitalized (I was wrong)

Then, re-evaluating:

* upstream reveal.js is never capitalized, it is actually proudly not-so, so this is accepted
* it is true that it is a converter on-top of asciidoctor so the space might represent that
* yet the command-line program is called with a dash...
* but without the dot...

`Asciidoctor-reveal.js` or `Asciidoctor reveal.js` what a harmless decision yet so difficult... I'll sleep on it.

Mogztter: A project has a name and an identifier. From what I've seen the rules are:

**Project name**
- Respect the upstream project capitalization
- Write Asciidoctor with a capital A
- Use a space between Asciidoctor and the upstream project

**Project identifier**
- If possible the ".js" extension should be dropped.
- The project identifier should be lower-kebab-case (ie. `asciidoctor-bespoke`, `asciidoctor-pdf`, `asciidoctor-epub3`).
- The project identifier should be used to name:
   * the binary
   * the project on GitHub
   * the Gem
   * the npm package

Here's an example:

**Asciidoctor Bespoke**
- Upstream project: Bespoke.js
- Project name: Asciidoctor Bespoke
- Project identifier: `asciidoctor-bespoke`
   * Gem: `asciidoctor-bespoke`
   * GitHub project: https://github.com/asciidoctor/asciidoctor-bespoke
   * Binary: `asciidoctor-bespoke`

References:
- [bin/asciidoctor-bespoke](https://github.com/asciidoctor/asciidoctor-bespoke/blob/d402d631c432bc641cce8d4a4f477c18ccafd758/bin/asciidoctor-bespoke#L14)
- [asciidoctor-bespoke.gemspec](https://github.com/asciidoctor/asciidoctor-bespoke/blob/d402d631c432bc641cce8d4a4f477c18ccafd758/asciidoctor-bespoke.gemspec#L5)
- [Bespoke.js](http://markdalgleish.com/projects/bespoke.js/)

Please note that the above rules also apply to:
- https://github.com/asciidoctor/asciidoctor-epub3
- https://github.com/asciidoctor/asciidoctor-pdf

Mogztter: > Asciidoctor-reveal.js or Asciidoctor reveal.js what a harmless decision yet so difficult... I'll sleep on it.

It should be: "Asciidoctor reveal.js using ". See:

- [bin/asciidoctor-bespoke](https://github.com/asciidoctor/asciidoctor-bespoke/blob/d402d631c432bc641cce8d4a4f477c18ccafd758/bin/asciidoctor-bespoke#L14)
- [bin/asciidoctor-epub3](https://github.com/asciidoctor/asciidoctor-epub3/blob/27b8f124020bf143e2ee7ce916f220a06fbadeb4/bin/asciidoctor-epub3#L15)
- [bin/asciidoctor-pdf](https://github.com/asciidoctor/asciidoctor-pdf/blob/f71116b0060ef896e84dd94f6d4126e6b87fea98/bin/asciidoctor-pdf#L15)

Mogztter: As a side note, we should probably update the Ruby binary [bin/asciidoctor-revealjs](https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/bin/asciidoctor-revealjs) to add the project name and version.

Mogztter: > what a harmless decision yet so difficult

Sorry I might have a small obsession on capitalization :laughing:

obilodeau: You've convinced me.

obilodeau:
> As a side note, we should probably update the Ruby binary bin/asciidoctor-revealjs to add the project name and version.

I got this to work (code from asciidoctor-pdf) but since I'm squash committing, I'm going to do a separate PR for this.

Mogztter:
Woot woot awesome work @obilodeau :rocket:

Issue #158, opened 2017-10-12
README should also point to stable releases documentation


obilodeau:
In addition to covering master, we should point to documentation for older releases like the 1.0.x train (maint-1.0.x branch) right at the top of the README.

See https://github.com/asciidoctor/jekyll-asciidoc/ README for an example. We might also want to point to the revealjs-2.0 branch also.

Idea from a comment in #157.

PR #317, opened 2020-01-22
Issue #158: Use section title (with sectnums)

eiswind:
I found that in helpers.rb the section_title was already defined with sectnum support.
I'm not sure if it is as simple as that, but I changed the section template to use this helper.

### DIFF
--- a/examples/level-sectnums.adoc
+++ b/examples/level-sectnums.adoc
@@ -1,0 +1,24 @@
+// .sections
+// Demonstration of section numbers and their specific reveal.js integration.
+// :include: //div[@class="slides"]
+// :header_footer:
+= Sectnums
+:backend: revealjs
+:sectnums:
+:sectnumlevels: 2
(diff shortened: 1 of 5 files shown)


Mogztter:
Seems to be working as expected but you will need to update the tests:

- `test/doctest/outline.html`
- `test/doctest/section.html`

See: https://travis-ci.org/asciidoctor/asciidoctor-reveal.js/jobs/640352964

eiswind:
Thanks for the hint :) There are tests...

obilodeau:
Looking good! Awesome that `helpers.rb` already provided that.

Can you create a simple AsciiDoc file in `examples/` that demonstrates the feature in a minimal way and with accompanying doctest.

I don't think README documentation is necessary since this is an upstream documented feature of the language that we will now support.

eiswind:
just added a simple example ..

PR #318, opened 2020-01-22
resolves #177 allow to override the built-in slim templates

### LINKED ISSUE #177: --template-dir argument doesn't seem to override the built-in slim templates
I'd like to use the `--template-dir` asciidoctor argument to change formatting produced by the default asciidoctor-revealjs slim templates. I've used this successfully with asciidoctor-bespoke, but when doing the same with asciidoctor-revealjs it doesn't seem to work (the backend template is not overridden). I'm thinking this might be due to #131 but would like to know if `-T` is supported or if there's a workaround that doesn't involve tinkering with the asciidoctor-revealjs backend.

thanks,

Mogztter:
resolves #177

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,6 +4,7 @@ require 'asciidoctor'
 require 'asciidoctor/doctest'
 require 'colorize'
 require 'tilt'
+require 'rake/testtask'
 
 CONVERTER_FILE = 'lib/asciidoctor-revealjs/converter.rb'
 JS_FILE = 'build/asciidoctor-reveal.js'
@@ -76,6 +77,7 @@ def build_converter(mode = :pretty)
         basebackend: 'html',
         outfilesuffix: '.html',
         filetype: 'html',
+        supports_templates: true
       },
       delegate_backend: 'html5',
       engine_opts: {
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -94,6 +96,11 @@ DocTest::RakeTasks.new do |t|
   t.converter_opts = { backend_name: 'revealjs' }
 end
 
+Rake::TestTask.new(:test) do |t|
```
Mogztter: I've created a test task to run tests in the `test/asciidoctor-revealjs` directory.
Previously, the existing file `test/asciidoctor-revealjs/version_test.rb` was not executed.

test/test_helper.rb
```diff
@@ -7,5 +7,4 @@
 
 require 'asciidoctor/doctest'
 require 'minitest/autorun'
-require 'minitest/rg'
```
Mogztter: This dependency is not explicitly declared (ie. missing)

Rakefile
```diff
@@ -132,7 +139,8 @@ namespace :examples do
   end
 end
 
+task 'test' => 'doctest'
 task 'doctest:test' => 'prepare-converter'
 task 'doctest:generate' => 'prepare-converter'
 # When no task specified, run test.
-task :default => :doctest
+task :default => :test
```
obilodeau: We should probably add `test` to travis' script so that the tests run on PR, etc.

Mogztter: Good idea and done!

Mogztter:
Thanks Olivier 🚀 🚀 🚀

PR #324, opened 2020-01-30
resolves #198 resolves docinfo head, header and footer

### LINKED ISSUE #198: docinfo.html is not inserted
Expected:
Including a docinfo.html file and a :docinfo: shared attribute should cause it to appear in the html output.

Actual:
HTML is not inserted.

Workaround:
I am able to `ln -s docinfo.html docinfo-header.html`
Then the html is inserted into head.
i.e.: file must be called docinfo-header.html to be inserted, which contradicts the documentation in asciidoctor user manual.

I discovered this workaround on observing that the code is looking for header here:
https://github.com/asciidoctor/asciidoctor-reveal.js/blob/6d700b903e52bd57e9aa9f8f477e1ecff859b7ab/templates/document.html.slim#L72

There is related and conflicting information here:
https://github.com/asciidoctor/asciidoctor/issues/1510

My repo demonstrating the problem:
https://github.com/timothypratley/enterprise-clojure-training

ggrossetie:
Inspired by https://github.com/asciidoctor/asciidoctor-bespoke/blob/master/templates/slim/document.html.slim

I've added an example to make sure that it's working and here the result:

tweet

As suggested by @mojavelinux I've added a unique suffix but it's a breaking change because `docinfo.html` won't be resolved (it must be renamed to `docinfo-revealjs.html`)

resolves #198

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 # test stuff
 /examples/*.html
+!/examples/docinfo*.html
 /examples/reveal.js/
 /examples/node_modules
 
--- a/README.adoc
+++ b/README.adoc
@@ -765,6 +765,41 @@ WARNING: This presentation is dangerous!
 
 Here are details about Asciidoctor's http://asciidoctor.org/docs/user-manual/#admonition-icons[Admonition icons] support.
 
+=== Supplemental Content with Docinfo
+// Originally from https://github.com/asciidoctor/asciidoctor-bespoke#supplemental-content
+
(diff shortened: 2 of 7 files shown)


obilodeau (approved):
After a careful review of this, including reading [bespoke's documentation](https://github.com/asciidoctor/asciidoctor-bespoke#supplemental-content) and [upstream API doc](https://github.com/asciidoctor/asciidoctor/blob/2acdcece395a8327373b5f7ebf01c71dcd80ec24/lib/asciidoctor/document.rb#L1023), I can say that I like it and it is going in.

I'll merge shortly but right before doing so, I'll add documentation and document the breaking behavior in the CHANGELOG.

PR #320, opened 2020-01-22
resolves #21 load highlight.js from a CDN

ggrossetie:
resolves #21 

It also resolves #319 even though it can be seen as a regression because the reveal.js highlight plugin contains all the languages by default.

To be backward compatible, we could use a local file `assets/highlight.min.js` that contains all the languages but I don't think it's a good solution (as you will rarely use all the 185 languages in a presentation).

- [x] add tests
- [x] make sure that the code is working using Node/JS

PR #327, opened 2020-02-08
Remove empty lines in the reveal.js dependencies configuration

ggrossetie:
Build an array and join the plugins in the `helpers.rb` file.
Another benefit is that we can now write unit tests to validate the logic.

### DIFF
--- a/templates/helpers.rb
+++ b/templates/helpers.rb
@@ -118,6 +118,21 @@ module Slim::Helpers
     end
   end
 
+  def revealjs_dependencies(document, node, revealjsdir)
+    dependencies = []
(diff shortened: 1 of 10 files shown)


obilodeau (approved):
That's good! If you have time, fix the conflict, if you don't, I'll handle it later.

templates/helpers.rb
```diff
@@ -118,6 +118,21 @@ def section_title(sec = self)
     end
   end
 
+  def revealjs_dependencies(document, node, revealjsdir)
+    dependencies = []
+    dependencies << "{ src: '#{revealjsdir}/plugin/highlight/highlight.js', async: true }" if (document.attr? 'source-highlighter', 'highlightjs')
```
obilodeau: Note to self: Now that #320 is in. This needs to be removed (Highlight.js is injected via docinfo) in addition to handle the conflict.

PR #328, opened 2020-02-09
Remove the last reference to reveal.js 3.7.0

ggrossetie:
Also use https://cdn.jsdelivr.net so we can use the latest version when it's released.


obilodeau (approved):
Great catch! Next time I'll grep for 3.9.2 instead of just diagonally reading ;)

PR #329, opened 2020-02-09
resolves #276 add the kdb style

### LINKED ISSUE #276: kbd macro doesn't work
Hi.

I'm' using asciidoctor 2.0.0

I'm using same source for html5, pdf and revealsjs for courses with a few adaptations depending the backend.

But I discoverd that kbd macro doesn't work with the revealjs backend.
It is not interpreted.
Do I miss something ?

ggrossetie:
Here's how it looks:

kbd

### DIFF
--- a/templates/asciidoctor-compatibility.css
+++ b/templates/asciidoctor-compatibility.css
@@ -37,6 +37,11 @@ thead{display:table-header-group}
 
 .reveal table.grid-none th,.reveal table.grid-none td{border-bottom:0!important}
 
+/* kbd macro */
+kbd{font-family:"Droid Sans Mono","DejaVu Sans Mono",monospace;display:inline-block;color:rgba(0,0,0,.8);font-size:.65em;line-height:1.45;background:#f7f7f7;border:1px solid #ccc;-webkit-border-radius:3px;border-radius:3px;-webkit-box-shadow:0 1px 0 rgba(0,0,0,.2),0 0 0 .1em white inset;box-shadow:0 1px 0 rgba(0,0,0,.2),0 0 0 .1em #fff inset;margin:0 .15em;padding:.2em .5em;vertical-align:middle;position:relative;top:-.1em;white-space:nowrap}
+.keyseq kbd:first-child{margin-left:0}
+.keyseq kbd:last-child{margin-right:0}
+
 /* callouts */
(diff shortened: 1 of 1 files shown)


templates/document.html.slim
```diff
@@ -23,7 +23,9 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
       link rel='stylesheet' href=(attr :revealjs_customtheme) id='theme'
     - else
       link rel='stylesheet' href='#{revealjsdir}/css/theme/#{attr 'revealjs_theme', 'black'}.css' id='theme'
-    include asciidoctor_revealjs.css.slim
+    /! This CSS is generated by the Asciidoctor reveal.js converter to further integrate AsciiDoc's existing semantic with reveal.js
+    style type="text/css"
+      include asciidoctor-default.css
```
obilodeau: It is not all just defaults. Maybe we should split into two files then. Upstream defaults and our overrides.

obilodeau (commented):
Refactoring with adding new functionality in a single commit makes it harder to assess if there's no regression silently introduced. Try to split both next time around.

I will diff the files with master once this is rebased to make sure nothing was reverted by mistake (since we touched the CSS along the way).

Mogztter:
> Refactoring with adding new functionality in a single commit makes it harder to assess if there's no regression silently introduced. Try to split both next time around.

I can split the changes no worries.

templates/document.html.slim
Mogztter: The name is misleading but I think we should keep everything in one file to avoid loading multiple files. Performance wise, I think it's better. But we can first add the default styles/rules taken from https://github.com/asciidoctor/asciidoctor/blob/master/data/stylesheets/asciidoctor-default.css and then our overrides?

obilodeau: I don't think performance is a strong argument. But I don't think having them separated is that better either. Let's go with what you propose.

obilodeau:
> I can split the changes no worries.

This was done in #330 and merged. Time to rebase :slightly_smiling_face:

templates/document.html.slim
Mogztter: > I don't think performance is a strong argument.

True :sweat_smile: 

> But I don't think having them separated is that better either. Let's go with what you propose.

Yes less is more and I don't see the value of two files :+1:

Mogztter:
@obilodeau Done! I've edited my first comment to add a screenshot.

obilodeau (changes requested):
Looking good!

Can you add that example in `examples/` and generate the appropriate doctest, please?

Mogztter:
> Can you add that example in examples/ and generate the appropriate doctest, please?

@obilodeau There's already an example and a doctest named "keyboard-shortcuts"

obilodeau (approved):
> @obilodeau There's already an example and a doctest named "keyboard-shortcuts"

Did I say something? :wink:

PR #330, opened 2020-02-10
Move the styles to a proper CSS file

ggrossetie:
It's mostly to be able to use features such as formatter/linter/minifier to make sure that the file is valid.


obilodeau:
This PR introduced a regression. I'm tracking it right now but I haven't found it yet.

Before
arjs-before

After
arjs-after

The color on the callouts was impacted

obilodeau:
It's the `background: rgba()` which is invalid (in chrome according to chrome dev tools) in the new form

templates/asciidoctor-compatibility.css
```diff
@@ -0,0 +1,50 @@
...
+.reveal table th.valign-middle,.reveal table td.valign-middle{vertical-align:middle}
+table thead th,table tfoot th{font-weight:bold}
+tbody tr th{display:table-cell;line-height:1.6}
+tbody tr th,tbody tr th p,tfoot tr th,tfoot tr th p{font-weight:bold}
+thead{display:table-header-group}
+
+.reveal table.grid-none th,.reveal table.grid-none td{border-bottom:0!important}
+
+/* callouts */
+.conum[data-value] {display:inline-block;color:#fff!important;rgba(0,0,0,.8);-webkit-border-radius:50%;border-radius:50%;text-align:center;font-size:.75em;width:1.67em;height:1.67em;line-height:1.67em;font-family:"Open Sans","DejaVu Sans",sans-serif;font-style:normal;font-weight:bold}
```
Mogztter: A word is missing here, it should be `;background:rgba(0,0,0,.8);` not `;rgba(0,0,0,.8);

obilodeau: Yeah just found it at the same time :+1:

PR #332, opened 2020-02-10
resolves #326 add grid layout

### LINKED ISSUE #326: Create a role for a two column layout
One of the common patterns in a presentation is to split things in two and have bullets on one side and {code, image, graph} on the other side.

I would like to provide CSS tied to a role to achieve that easily.

* Must allow combining several block types: image with code, bullets with bullets, logos, bullets with code, etc.
* Could allow setting a specific background color (#234)
* Don't break background compatibility (that said we are not responsible of avoiding downstream CSS breakage)

See what others have done:

### Asciidoctor PDF.js

[CSS source](https://github.com/Mogztter/asciidoctor-pdf.js/blob/master/examples/slides/slides.css)
```css
.two-col > .slide-content {
  display: flex;
  flex-direction: row;
}

.two-col > .slide-content > *:first-child {
  margin-right: 10px;
  flex: 50%;
}
.two-col > .slide-content > *:last-child {
  margin-left: 10px;
  flex: 50%;
}
```

### A user presentation

Markup ([source](https://github.com/bentolor/java9to13/blob/master/docs/presentation-language.adoc)):
```AsciiDoc
=== Local Variable Type Inference [jep]#https://openjdk.java.net/jeps/286[JEP 286]#

[.col2]
--
* New reserved Java type `var`
* _„real type“_ inferred by compiler *on first initialization*
* can denote classes (`Integer`) & primitives (`int`)
* *only*  available for _local variables_
--

[.source.col2,java]
----
var primitiveVal = 5;   // int
var doubleVal    = 5d;  // double

final var sl = new ArrayList<String>();

var letters = "";
for (var s : sl) {
    letters += s.toLowerCase();
}

// var: enable annotation w/ type infer
(var x, @NotNull var y) -> x + y;
----
```

CSS ([source](https://github.com/bentolor/java9to13/blob/master/docs/css/presentation.css#L269)):
```css
.col2 {
  width:48% !important;
  margin: 0 1% !important;
  float:left;
}
/* Aufzählung nicht zentrieren im 2col layout.*/
.reveal .col2 ol, .reveal .col2  dl, .reveal .col2  ul {
    display:table;
}
```

Mogztter:
resolves #326

2colums

3columns

columns-size

multi-column-wrap

column-background

### DIFF
--- a/examples/grid-layout.adoc
+++ b/examples/grid-layout.adoc
@@ -1,0 +1,100 @@
+= Grid Layout
+:docinfo: private
+:source-highlighter: highlight.js
+:highlightjs-languages: clojure,scala
+
+[.columns]
+== 2 columns
+
+[.column]
+--
+* **Edgar Allen Poe**
+* Sheri S. Tepper
+* Bill Bryson
+--
+
+[.column]
+--
+Edgar Allan Poe (/poʊ/; born Edgar Poe; January 19, 1809 – October 7, 1849) was an American writer, editor, and literary critic.
+--
+
+[.columns.wrap]
+== Multi-columns with wrap
+
+[.column]
+--
+.Kotlin
+[source,kotlin]
+----
+fun main() {
+    println("Hello, World!")
+}
+----
+--
+
+[.column]
+--
+.Clojure
+[source,clojure]
+----
+(defn -main [& args]
+  (println "Hello, World!"))
+----
+--
+
+[.column]
+--
+.Scala
+[source,scala]
+----
+object HelloWorld {
+  def main(args: Array[String]): Unit = {
+    println("Hello, world!")
+  }
+}
+----
+--
+
+[.columns]
+== Columns with size
+
+[.column.is-one-third]
+--
+* **Kotlin**
+* Java
+* Scala
+--
+
+[.column]
+--
+Programming language for Android, mobile cross-platform
+and web development, server-side, native,
+and data science. Open source forever Github.
+--
+
+[.columns]
+== 3 columns
+
+[.column]
+* Java
+* **Kotlin**
+
+[.column]
+* Node
+* **Deno**
+
+[.column]
+* Ruby
+* **Crystal**
+
+
+[.columns.is-half.green.left]
+== !
+
+[.column.is-half]
+* Java
+* **Kotlin**
+
+[.column.is-half]
+* Ruby
+* **Crystal**
--- a/examples/grid-layout.adoc
+++ b/examples/grid-layout.adoc
@@ -2,6 +2,7 @@
 :docinfo: private
 :source-highlighter: highlight.js
 :highlightjs-languages: clojure,scala
+:revealjs_hash: true
 
 [.columns]
 == 2 columns
@@ -98,3 +99,40 @@ and data science. Open source forever Github.
 [.column.is-half]
 * Ruby
 * **Crystal**
+
+== Languages
+
+We will go over languages
+
+[.columns]
+=== Kotlin
+
+[.column]
+[source,kotlin]
+----
+fun main() {
+    println("Hello, World!")
+}
+----
+
+[.column]
+--
+* Nice and compact
+* Works well on Android
+--
+
+[.columns]
+=== Clojure
+
+[.column]
+[source,clojure]
+----
+(defn -main [& args]
+  (println "Hello, World!"))
+----
+
+[.column]
+--
+* Lisp-like
+* Few simple concepts
+--
--- a/templates/asciidoctor-compatibility.css
+++ b/templates/asciidoctor-compatibility.css
@@ -68,3 +68,70 @@ td.hdlist1{font-weight:bold;padding-bottom:1.25em}
 .reveal .colist>table th, .reveal .colist>table td {border-bottom:0}
 /* Fixes line height with Highlight.js source listing when linenums enabled (see #331) */
 .reveal .hljs table thead tr th, .reveal .hljs table tfoot tr th, .reveal .hljs table tbody tr td, .reveal .hljs table tr td, .reveal .hljs table tfoot tr td{line-height:inherit}
+
+/* Columns layout */
+.columns .slide-content {
+  display: flex;
+}
+
+.columns.wrap .slide-content {
+  flex-wrap: wrap;
+}
+
+.columns .slide-content > .column {
+  display: block;
+  flex-basis: 0;
+  flex-grow: 1;
+  flex-shrink: 1;
+  padding: .75rem;
+}
+
+.columns .slide-content > .column.is-full {
+  flex: none;
+  width: 100%;
+}
+
+.columns .slide-content > .column.is-four-fifths {
+  flex: none;
+  width: 80%;
+}
+
+.columns .slide-content > .column.is-three-quarters {
+  flex: none;
+  width: 75%;
+}
+
+.columns .slide-content > .column.is-two-thirds {
+  flex: none;
+  width: 66.6666%;
+}
+
+.columns .slide-content > .column.is-three-fifths {
+  flex: none;
+  width: 60%;
+}
+
+.columns .slide-content > .column.is-half {
+  flex: none;
+  width: 50%;
+}
+
+.columns .slide-content > .column.is-two-fifths {
+  flex: none;
+  width: 40%;
+}
+
+.columns .slide-content > .column.is-one-third {
+  flex: none;
+  width: 33.3333%;
+}
+
+.columns .slide-content > .column.is-one-quarter {
+  flex: none;
+  width: 25%;
+}
+
+.columns .slide-content > .column.is-one-fifth {
+  flex: none;
+  width: 20%;
+}
--- a/templates/section.html.slim
+++ b/templates/section.html.slim
@@ -54,11 +54,12 @@
     data-state=(attr 'state'))
     - unless hide_title
       h2=section_title
-    - if parent_section_with_vertical_slides
-      - (blocks - vertical_slides).each do |block|
-        =block.convert
-    - else
-      =content.chomp
+    div.slide-content
+      - if parent_section_with_vertical_slides
+        - (blocks - vertical_slides).each do |block|
+          =block.convert
+      - else
+        =content.chomp
 
 / RENDERING
 / render parent section of vertical slides set
--- a/templates/section.html.slim
+++ b/templates/section.html.slim
@@ -54,12 +54,15 @@
     data-state=(attr 'state'))
     - unless hide_title
       h2=section_title
-    div.slide-content
-      - if parent_section_with_vertical_slides
-        - (blocks - vertical_slides).each do |block|
-          =block.convert
-      - else
-        =content.chomp
+    - if parent_section_with_vertical_slides
+      - unless (_blocks = blocks - vertical_slides).empty?
+        div.slide-content
+          - _blocks.each do |block|
+            =block.convert
+    - else
+      - unless (_content = content.chomp).empty?
+        div.slide-content
+          =_content
 
 / RENDERING
 / render parent section of vertical slides set
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 # test stuff
 /examples/*.html
 !/examples/docinfo*.html
+!/examples/*docinfo-revealjs.html
 /examples/reveal.js/
 /examples/node_modules
 
--- a/README.adoc
+++ b/README.adoc
@@ -21,7 +21,7 @@ endif::[]
 :uri-revealjs-home: http://lab.hakim.se/reveal-js/
 :uri-revealjs-gh: https://github.com/hakimel/reveal.js
 :uri-nodejs-download: https://nodejs.org/en/download/
-:showcasedir: showcase 
+:showcasedir: showcase
 
 ifdef::env-github[]
 image:https://travis-ci.org/asciidoctor/asciidoctor-reveal.js.svg?branch=master[Build Status,link=https://travis-ci.org/asciidoctor/asciidoctor-reveal.js]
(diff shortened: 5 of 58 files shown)


templates/asciidoctor-compatibility.css
```diff
@@ -53,3 +53,70 @@ b.conum *{color:inherit!important}
 
 /* Override Asciidoctor CSS that causes issues with reveal.js features */
 .reveal .hljs table{border: 0}
+
+/* Columns layout */
+.columns .slide-content {
```
Mogztter: Not sure if we should (manually) minify the content here? (ie. remove the extra spaces, the trailing semi-colon and remove line breaks)

templates/section.html.slim
```diff
@@ -57,8 +57,9 @@
 
       - unless hide_title
         h2=section_title
-      - (blocks - vertical_slides).each do |block|
-        =block.convert
+      div.slide-content
+        - (blocks - vertical_slides).each do |block|
+          =block.convert
     - vertical_slides.each do |subsection|
```
Mogztter: I'm a bit confused with this part of the code. I need/want to wrap the content of the slide (everything except the title in a `<div class="slide-content">`.

Mogztter:
@obilodeau I will fix the doctests but first I want to make sure that you are OK with the changes.

templates/asciidoctor-compatibility.css
obilodeau: I'm not fond of aggressive minification. This is for slide decks with heavy images and videos. Shaving a few bytes here isn't going to make much of a difference.

OTOH, DevTools now have prettyfication tools built-in and it's easy to see what CSS is applied where.

I don't have a strong opinion on either. However, we did minify some of the existing CSS already. Your call.

templates/section.html.slim
obilodeau: The way I see it: vertical slides are just an organization mechanism. They are full slides in their own right so the `slide-content` div should apply to them individually, not to the vertical set. You seem to be doing the right thing here.

obilodeau:
I'm going to have to give this a serious try before I can conclude. With existing decks where I wish I had that.

Try adding vertical slides to your example too, just to validate:

```AsciiDoc
== Languages

We will go over languages

=== Java
// TODO code snippet on the left, bullets saying why it's awesome on the right

=== Kotlin
// TODO code snippet on the left, bullets saying why it's awesome on the right
```

examples/grid-layout.adoc
```diff
@@ -0,0 +1,100 @@
...
+[.column]
+* Node
+* **Deno**
+
+[.column]
+* Ruby
+* **Crystal**
+
+
+[.columns.is-half.green.left]
```
obilodeau: The green doesn't work right now. It's missing something.

I remember background coloring being a bit tricky to get working because they need to be applied to an entirely different part of the DOM. See `examples/background-color.adoc`. Can you make sure your example works please?

obilodeau:
* Merged with master to integrate #333 changes and resolve conflict once
* Added vertical slides example
* Found a bug in my merge where div.slide-content was not created on the top of a vertical slides series

I'm still testing decks with it

templates/asciidoctor-compatibility.css
Mogztter: I don't think we should do it.

examples/grid-layout.adoc
Mogztter: Ooops the docinfo was ignored. I've modified the `.gitignore` file and pushed `examples/grid-layout-docinfo-revealjs.html`

examples/grid-layout.adoc
```diff
@@ -0,0 +1,138 @@
...
+[source,clojure]
+----
+(defn -main [& args]
+  (println "Hello, World!"))
+----
+
+[.column]
+--
+* Lisp-like
+* Not sure why people like it
```
Mogztter: Are you trolling Clojure? :laughing:

Mogztter:
> Found a bug in my merge where div.slide-content was not created on the top of a vertical slides series

Did you fix it?

> I'm still testing decks with it

Let me know if it looks fine, I will update the doctests to add the wrapping `<div class="slide-content">`.

examples/grid-layout.adoc
obilodeau: Yeah, you should probably change that. I didn't know what to say about Clojure and I wanted two bullets. I haven't done Java or any JVM-language for 10+ years.

obilodeau:
> Did you fix it?

Yes I did in 275e91e. Sorry if it wasn't clear.

obilodeau:
There is a regression in `examples/video.adoc`.

Before change:
adrjs-cols-video-works

After change:
adrjs-cols-video-regression

Mogztter:
Nice catch! 
I will try to add visual tests to catch regressions more easily.

obilodeau:
After some messing around, it is clear that it is the `stretch` class that causes problems. Slide 3 of `examples/source-callouts.html` also has a problem.

`stretch` is special, in the sense that it's not changing how things look with CSS. It relies on special reveal.js JavaScript. You can read about the feature here: https://github.com/hakimel/reveal.js/#stretching-elements

Note:
>Limitations:
>* Only direct descendants of a slide section can be stretched
>* Only one descendant per slide section can be stretched

From past observation, I would say that it injects maximum width and height in pixels directly in the `style=` HTML attribute of the HTML tag where the `stretch` class is present.

It is applied by default to videos because who would want a small video box in a slide.

Mogztter:
We should probably recreate this feature using CSS or workaround the "only direct descendant" limitation.

Mogztter:
Using flexbox "everywhere" could add more flexibility but it does require a few changes and probably a few breaking changes since stretch/grow need to be on the parent (if we want to use CSS and not JavaScript).

Anyway, I will try to fix this issue with the least amount of work. We can then decide if we want to switch to flexbox/grid or stick with the absolute/center positioning.

Mogztter:
stretch element sizing is done in the [layoutSlideContents](https://github.com/hakimel/reveal.js/blob/94d98ff7d89e35c7d5b289ba387cdb0d42b30793/js/reveal.js#L2362) function which is called in the `layout` function which is called when the window is resized.

But `layout` function is also called when the overview is activated/deactivated or when we go to the next/previous slide. So I think that [registering a listener](https://github.com/hakimel/reveal.js#resize-event) to size the nested stretch elements won't be enough 🤔

obilodeau:
I'm surprised that this problem exists. I mean no one ever used two-columns layouts in reveal.js!?

Can it be done as a plugin instead of monkey-patching? That would work as well.

Mogztter:
> I'm surprised that this problem exists. I mean no one ever used two-columns layouts in reveal.js!?

It can definitely be done without using a wrapping `<div>` (and flexbox) but it will be a bit "limited" in comparison. 
I think we should implement this feature without using flexbox. I don't want to rush a half-baked implementation based on flexbox.
Please note that switching to flexbox will most likely require a major version (ie. 5.x)

> Can it be done as a plugin instead of monkey-patching? That would work as well.

Maybe.

obilodeau:
I'm not sure about going without it... While we are here shouldn't we get this right? Besides the `stretch` issue this seems like a great way to solve the basic layouts for presentations fairly easily.

Regarding the stretch issue, we control everything. Isn't there a way to maximize a single flex div? We put the stretch attribute on videos and we can even intercept the attribute and do something else.

Even bolder, couldn't we add an `onChange()` event on the big div and propagate the values set by reveal.js to the first stretch child? Brainstorming here

Mogztter:
Let me try to summarize.
The main issue is that `height: 100%` does not work.
I believe it's because the parent has an absolute positioning and/or because reveal.js is dynamically updating the height, width and zoom on the parent.
Anyway, `height:100%` on the `<section>` element is working as expected. But does not work on child elements.

Let's take an example:

**Simplified HTML**
```html
<body>
  <div class="reveal slide">
    <div class="slides" style="width: 960px; height: 700px; scale(1);">
      <section>                                    <!-- height:100% = 700px + 40px (padding) = OK -->
        <h2>Stretched Callout</h2>
        <div class="slide-content">                <!-- height:100% = 700px = KO -->
          <div class="listingblock stretch">       <!-- height:100% = 700px = KO -->
          </div>
          <div class="colist arabic">
          </div>
        </div>
      </section>
    </div>
  </div>
</body>
```

Here, the element "slide-content" has a 100% height of 700px but it's wrong because it ignores the `<h2>` element height. Even worse, the "listingblock" element also has a 100% height of 700px. In this particular case, it means that the "colist" element will be pushed outside of the slide :scream: 

> Regarding the stretch issue, we control everything. Isn't there a way to maximize a single flex div? We put the stretch attribute on videos and we can even intercept the attribute and do something else.

Yes we can.
To workaround this issue, I can use [flex-grow](https://developer.mozilla.org/fr/docs/Web/CSS/flex-grow) but as far as I know I need to apply the `flex-grow: 1` on each parent:

* `<section>` = height:100% + flex-grow:1
* `<div class="slide-content">` = flex-grow:1
* `<div class="listingblock stretch">` = flex-grow:1

The main issue is that we need to use `display: flex` on the section. This value is controlled by a reveal.js option named `display`.
At this point, I thought, let's just ignore this value and just use `display: flex` everywhere, see: https://github.com/asciidoctor/asciidoctor-reveal.js/compare/master...Mogztter:issue-326-columns-layout-flex
It's working but it introduces a few changes as it does not rely on JavaScript anymore. I guess we could use a bit of JavaScript to propagate the stretch value on the parent elements. For now, I'm using a `grow` role on the slide.

obilodeau:
Does it negatively affects?
* Speaker mode (press `s`)
* Overview mode (press `Esc`)
* PDF output (add `?print-pdf` to the URL and print without margins)

I wonder why the `config.display` feature of reveal.js is there...

Mogztter:
> Does it negatively affects?

Not sure but since we need JavaScript to compute the max height, I will try to workaround this limitation "Only direct descendants of a slide section can be stretched". 

> I wonder why the config.display feature of reveal.js is there...

I believe that @mojavelinux added it to configure a flex display. The `display` value is set (via JavaScript) on the `section` element, so it adds flexibility.
In my branch, I force the value to "flex": https://github.com/asciidoctor/asciidoctor-reveal.js/compare/master...Mogztter:issue-326-columns-layout-flex#diff-38f9a53261d123f9921bdc0090cc4912R224

obilodeau:
>> I wonder why the config.display feature of reveal.js is there...
>
> I believe that @mojavelinux added it to configure a flex display. The `display` value is set (via JavaScript) on the `section` element, so it adds flexibility.
In my branch, I force the value to "flex": https://github.com/asciidoctor/asciidoctor-reveal.js/compare/master...Mogztter:issue-326-columns-layout-flex#diff-38f9a53261d123f9921bdc0090cc4912R224

*/me did a `git blame` upstream.*

Oh my god, wow! This change really was introduced by @mojavelinux. Probably before he started using bespoke :open_mouth: I didn't know.

Mogztter:
It's fairly easy to remove the limitation so my guess is that the decision was made for performance reasons. I'm asking upstream about a possible solution.

Mogztter:
Can you please give it a try @obilodeau?

obilodeau:
I'll try later tonight. Thanks!

obilodeau:
## Video

On the first load of the converted `examples/video.html`, when I get to the vimeo slide there's a glitch:

adrjs-cols-video-glitch-before-resize

There's an empty space at the top. On other videos this pushes the video slightly out of the screen. I will add a test case in this deck where this is worse. Using the embedded reveal.js video player.

If I manually resize the browser by dragging around the window then replacing it full screen the glitch is gone:

adrjs-cols-video-glitch-after-resize

I was able to confirm the same glitch on a stretched image in a private slide deck.

## Misc

I noticed that YouTube embeds didn't autoplay but I had the same problem with master so we can ignore for the sake of this PR.

## Results

Besides the above glitch, I tested several decks and they were fine! :+1:

obilodeau:
Here's the other example. Available with: `examples/video.html#/_remote_video_file_auto_sized`

Glitched:

adrjs-cols-video-glitch-before-resize-2


After a manual resize, not glitched:

adrjs-cols-video-glitch-after-resize-2

Hopefully you'll be able to track this one down. Good luck!

Mogztter:
I thought it was unnecessary to resize the stretch elements when the slide changed but I guess I was wrong...
I've just added an event listener, it seems to fix this issue.

obilodeau:
Great!

When testing, I thought I found a problem with PDF rendering on a large slide deck, so I attempted to make a simpler case where it is reproduced. However, in my simpler case, I couldn't reproduce.

However, I found the problem.

It boils down to the theme that I made. I still use this CSS from another theme:
```css
/* Ensure certain elements are never larger than the slide itself */
.reveal img,
.reveal video,
.reveal iframe {
  max-width: 95%;
  max-height: 95%;
  }
```

But if I maximize the height of an image (with `.stretch`), it gets slightly off (even with 95% max) so it gets bumped out in print mode. Reducing this to 90% in DevTools fixed it. This is probably due to the font and the spacing on top in my theme which requires this change. Anyway, long story short, the issue seems to be in my theme.

To me this seems to be a :+1:. How do you feel about this PR @Mogztter?

Mogztter:
> But if I maximize the height of an image (with .stretch), it gets slightly off (even with 95% max) so it gets bumped out in print mode. Reducing this to 90% in DevTools fixed it. This is probably due to the font and the spacing on top in my theme which requires this change. Anyway, long story short, the issue seems to be in my theme.

It sounds like a limitation of the original `getRemainingHeight` function that do not or cannot take everything into account.

> To me this seems to be a +1. 

Should I rebase on master?

> How do you feel about this PR @Mogztter?

I'm not 100% satisfied because we still need to use JavaScript but I think it's the best we can do for now. In my opinion, the column definition is quite simple and will probably be enough in the majority of cases. So I think we should go for it!

obilodeau:
Thanks by the way! This is something I wanted for a while and I wasn't sure how to tackle it in a flexible manner. Pun intended.

A very nice piece of work. Hopefully, upstream will be kind to us and we will be able to remove that JavaScript patch.

Mogztter:
Should I remove the `col2` role in the _source-emphasis.html_ example?

> Add doctest and this is good to go!

I found a few issues while updating doctests :wink: 

> You can document the feature in the README if you are willing but if you don't I'll do it.

Ok!

> Thanks by the way! This is something I wanted for a while and I wasn't sure how to tackle it in a flexible manner. Pun intended.

:blush: 

> A very nice piece of work. Hopefully, upstream will be kind to us and we will be able to remove that JavaScript patch.

Yep, that would be nice :+1:

obilodeau:
> Should I remove the `col2` role in the *source-emphasis.html* example?

Yes, replace it with this new stuff and drop its `customcss` file.

examples/grid-layout.adoc
```diff
@@ -0,0 +1,138 @@
...
+[source,clojure]
+----
+(defn -main [& args]
+  (println "Hello, World!"))
+----
+
+[.column]
+--
+* Lisp-like
+* Few simple concepts
```
obilodeau: :+1:

Mogztter:
Let's go :rocket: :rocket: :rocket:

PR #333, opened 2020-02-10
Use content_for and yield_content to remove code duplication

ggrossetie:


obilodeau (approved):
Oh wow, this is neat! Pretty compact and clever code that I wouldn't have been able to write myself.

PR #340, opened 2020-02-17
Added is-vcentered to our columns options

obilodeau:
It's a lot prettier when columns are unbalanced.

[Directly inspired by Bulma](https://bulma.io/documentation/columns/options/), good thing you mentioned it in the README @Mogztter!

### DIFF
--- a/examples/grid-layout.adoc
+++ b/examples/grid-layout.adoc
@@ -1,3 +1,7 @@
+// .grid-layout
+// Testing our columns roles
+// :include: //div[@class="slides"]
+// :header_footer:
 = Grid Layout
 :docinfo: private
 :source-highlighter: highlight.js
@@ -135,4 +139,36 @@ fun main() {
 --
 * Lisp-like
 * Few simple concepts
+--
+
+[.columns.is-vcentered]
+== Vertically Centered Columns
+
+[.column]
+* Java
+* **Kotlin**
+
+[.column]
+* Node
+* **Deno**
+
+[.column.is-half]
+--
+.Source
+[source,asciidoc]
+....
+[.columns.is-vcentered]
+== Vertically Aligned Columns
+
+[.column]
+* Java
+* **Kotlin**
+
+[.column]
+* Node
+* **Deno**
+
+[.column.is-half]
+[...]
+....
 --
--- a/templates/asciidoctor-compatibility.css
+++ b/templates/asciidoctor-compatibility.css
@@ -78,6 +78,10 @@ td.hdlist1{font-weight:bold;padding-bottom:1.25em}
   flex-wrap: wrap;
 }
 
+.columns.is-vcentered .slide-content {
+  align-items: center;
+}
+
 .columns .slide-content > .column {
   display: block;
   flex-basis: 0;
--- a/test/doctest/grid-layout.html
+++ b/test/doctest/grid-layout.html
@@ -1,0 +1,273 @@
+<!-- .grid-layout -->
+<div class="slides">
+  <section class="title" data-state="title">
+    <h1>Grid Layout</h1>
+  </section>
+  <section class="columns" id="_2_columns">
+    <h2>2 columns</h2>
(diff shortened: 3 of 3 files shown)


Mogztter:
I had a pull request in the making with `is-vcentered` but using a different use case.
I was using the following document:

```adoc
= Alignments
:revealjs_center: false
:revealjs_display: flex

[.is-vcentered]
== Vertical alignment

Here's a list of writers:

* Edgar Allen Poe
* Sheri S. Tepper
* Bill Bryson

// not sure about the name "is-vcentered-slide"
[.is-vcentered-slide]
== Vertical alignment (slide)

Here's a list of writers:

* Edgar Allen Poe
* Sheri S. Tepper
* Bill Bryson
```

In fact, there are multiple ways to vertically align content.

In the first slide, the space above the content (and the bottom of the title) and the space below the content (and the bottom of the slide) are equals:

valign1

In the second slide, the space above the top of the content (and the top of the title) and the space below the bottom of the content (and the bottom of the slide) are equals (ignoring the title height):

valign-slide2

And we can also vertically align content depending on the tallest column (as you did):
valign-column


It's somehow related because I want to make sure that the naming will be consistent and that we can support different alignements.
I'm not a big fan of `.is-vcentered-slide` maybe we should add an extra role on the content itself?

```
[.is-vcentered]
== Vertical alignment (slide)

[.absolute]
--
Here's a list of writers:

* Edgar Allen Poe
* Sheri S. Tepper
* Bill Bryson
--
```

obilodeau:
> I'm not a big fan of `.is-vcentered-slide` maybe we should add an extra role on the content itself?

Why not just `.vcentered`? The `.is-vcentered` is a modifier on something that exists (`.columns` as in "the columns are vertically centered") but here we don't modify anything. I'm not convinced by the way. It's just a thought.

Mogztter:
> Why not just .vcentered? The .is-vcentered is a modifier on something that exists (.columns as in "the columns are vertically centered")

I think we should stick with Bulma naming. They (hopefully) did some researches 😄
I was talking about how to make a distinction between "vertically align the content (within the content space)" and "vertically align the content within the whole slide".
I think `is-absolute` can work because that's basically what we are doing (ie. applies `position: absolute`) and it's less frequent use case (and it has a few drawbacks, for instance the content can overlap with the title since the position is absolute).

> but here we don't modify anything. I'm not convinced by the way. It's just a thought.

We do modify the alignment, right?

I think we can merge this pull request, I can follow-up with another pull request to implement "vertically align the content (within the content space)".
Since `is-absolute` is an edge case we can leave it for now.

obilodeau:
> I think we should stick with Bulma naming. They (hopefully) did some researches 😄

Yes, of course, that comment was not about replacing `is-vcentered` it was about finding an alternative to `is-vcentered-slide`.

> I think `is-absolute` can work because that's basically what we are doing (ie. applies `position: absolute`) and it's less frequent use case (and it has a few drawbacks, for instance the content can overlap with the title since the position is absolute).

To me, this is too close to implementation detail. `is-absolute` doesn't give me a clue that the content is going to be vertically centered. I think we should try other things.

>> but here we don't modify anything. I'm not convinced by the way. It's just a thought.
>
> We do modify the alignment, right?

Of course :slightly_smiling_face:. I meant `.is-vcentered-slide` is not modifying *anything else* so we don't need to add the *is* on top of something existing. I think bulma uses `.is-` when it's an option on top of some existing base component: `.columns.is-vcentered`, `.column.is-one-third`, having `.is-one-third` alone is incomplete. One-third of what? `one-third-width` would make more sense if it's not an option on top of something else.

Just replying here while we are discussing this but I agree that this should be done in a follow-up PR.

> I think we can merge this pull request, I can follow-up with another pull request to implement "vertically align the content (within the content space)".
> Since `is-absolute` is an edge case we can leave it for now.

:+1: I'll merge later tonight (EST) then release.

Issue #144, opened 2017-09-22
asciidoctor-doctest invoked from the javascript toolchain


obilodeau:
doctest is really useful to track regressions anywhere in the asciidoc stack (us, upstream or dependencies). However, right now they are only invoked from the ruby code and thus problems introduced by other components (like Opal) can't be detected by it currently.

asciidoctor-doctest is working on allowing its test suite to be run from the command-line, allowing any interpreter. We would then integrate that into our Travis automated testing.

Ref: asciidoctor/asciidoctor-doctest#13

Mogztter:
I've opened a pull request to make the processor configurable in asciidoctor-doctest but unfortunately it doesn't work yet.

Could you take a look ?

https://github.com/asciidoctor/asciidoctor-doctest/pull/16

Once this is fixed we could implement a custom processor to use Asciidoctor.js CLI in asciidoctor-doctest :tada:

ggrossetie:
This landed, just via a different (and better) route than the asciidoctor-doctest PR discussed above — that gem was dropped entirely in favor of [asciidoc-testkit](https://github.com/ggrossetie/asciidoc-testkit).

`rake testkit:test:js` now runs the exact same fixture corpus (the shared 37 generic AsciiDoc-construct families, plus this project's own `standalone` presentations) through the **native JavaScript converter** (Asciidoctor.js, no Opal) and checks it against the same expected HTML the Ruby converter is held to — this is exactly "invoke the doctest-style suite from any interpreter" from the issue description, and it runs in CI on every change.

Closing as resolved by the migration to asciidoc-testkit.

PR #343, opened 2020-02-18
resolves #342 replace pattern with generated code (no substitution)

### LINKED ISSUE #342: Broken syntax highlighting on master with node
When `lib/asciidoctor-revealjs/highlightjs.rb`'s `HIGHLIGHT_PLUGIN_SOURCE` is converted, there's a `\n` somewhere in that string that gets transformed into a newline in the destination HTML resulting in a broken line of javascript.

Error console:
adrjs-node-js-newline

Ruby code with the `\n`:
adrjs-node-js-newline-ruby-ok

Converted to HTML with a newline (instead of the `\n`):
adrjs-node-js-newline-html-not-ok

Mogztter:
resolves #342

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -40,7 +40,8 @@ namespace :build do
     mkdir_p [File.dirname(JS_FILE), File.dirname(DIST_FILE)]
     File.open(JS_FILE, 'w') do |file|
       template = File.read('src/asciidoctor-revealjs.tmpl.js')
-      file << template.sub('//OPAL-GENERATED-CODE//', builder.to_s)
+      template['//OPAL-GENERATED-CODE//'] = builder.to_s
+      file << template
     end
     File.binwrite "#{JS_FILE}.map", builder.source_map
 


obilodeau:
It would be nice to have something failing Travis that is fixed by that commit. To avoid regressions.

Mogztter:
If we want to make sure that client-side syntax highlighting is working then we need to open the HTML page in a browser and check that code block is highlighted.
If you want I can add an "integration" test based on Puppeteer? We could also use this approach when slides are generated using Ruby.

Another option is to invoke asciidoctor-doctest from the JavaScript toolchain: https://github.com/asciidoctor/asciidoctor-reveal.js/issues/144
I did give it a try last year but I couldn't figure out how to plug Asciidoctor.js...

obilodeau:
### Testing strategy

Another would be to diff the resulting converted examples from node to the ruby ones.

I did an initial attempt at this:
```diff
$ wdiff -3 examples/release-4.0.html release-4.0.html

======================================================================
 [-charset="utf-8" /><meta-] {+charset="utf-8"><meta+}
======================================================================
 [-minimal-ui" /><title>Asciidoctor-] {+minimal-ui"><title>Asciidoctor+}
======================================================================
 [-href="reveal.js/css/reset.css" /><link-] {+href="reveal.js/css/reset.css"><link+}
======================================================================
 [-href="reveal.js/css/reveal.css" /><link-] {+href="reveal.js/css/reveal.css"><link+}
======================================================================
 [-id="theme" /><!--This-] {+id="theme"><!--This+}
======================================================================
 [-href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.0-1/css/all.min.css" /><link-] {+href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.0-1/css/all.min.css"><link+}
======================================================================
 [-href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.0-1/css/v4-shims.min.css" /><!--Printing-] {+href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.0-1/css/v4-shims.min.css"><!--Printing+}
======================================================================
 [-href="release-4.0.css" /></head><body><div-] {+href="release-4.0.css"></head><body><div+}
======================================================================
 [-logo" /></div>-] {+logo"></div>+}
======================================================================
```

The only differences are how some tags are closed differently. Being pragmatic, until we have proper doctest, this can be removed:

```
$ diff -u <(sed 's/ //g;s/\///g' examples/release-4.0.html) <(sed 's/ //g;s/\///g' release-4.0.html) 
$ echo $?
0
```

Do you think this idea is worth pursuing? I don't have time to pursue a Puppeteer approach nor implement doctest for the Javascript toolchain. But a rake task that does this seems easy to achieve and would have caught that bug.

Anyways, I created issue #345.

### Back to the PR

I tried it, it works. I'm going to merge this now.

Mogztter:
> Being pragmatic, until we have proper doctest...

It's a good compromise for now :+1:

Issue #348, opened 2020-02-24
Cannot use MathJax locally, only via CDN


akosma:
Currently there is no way to specify that the MathJax library should be loaded locally, instead of from a CDN. This would be akin to using the `:revealjsdir:` attribute that specifies a location for the library, for example within the `node_modules` folder.

In the source of `document.html.slim` there's a `cdn_base` variable [set](https://github.com/asciidoctor/asciidoctor-reveal.js/blob/8ad7fde2f23e0a36281a0b0d1721722b98e6f4ec/templates/document.html.slim#L15), which is [used](https://github.com/asciidoctor/asciidoctor-reveal.js/blob/8ad7fde2f23e0a36281a0b0d1721722b98e6f4ec/templates/document.html.slim#L55) for the URL to MathJax, and it cannot be overridden by an attribute right now.

A variable such as `:mathjaxdir:` or similar could specify the location of the library, either locally or in another server, instead of just a CDN.

obilodeau:
### Example required

Can you provide a one-slide AsciiDoc document with MathJax in it, please? I realize we don't have anything in `examples/` and I wouldn't know what to write myself.

### DIY

Supporting that feature should be relatively straightforward, do not hesitate if you want to attempt it. `HACKING.adoc` has documentation on how to interact with templates when trying to implement something (pry byebug REPL).

### reveal.js support

Lastly, reveal.js also [has MathJax support](https://github.com/hakimel/reveal.js#mathjax) on its own.  It looks like you can specify a path to `MathJax.js` in a configuration block. You can try it with reveal.js by using our custom plugin and plugin customization features.

akosma:
Please find attached a simple possible project to demonstrate the issue: [issue348.zip](https://github.com/asciidoctor/asciidoctor-reveal.js/files/4249105/issue348.zip). You can use the `make` command to rebuild the `slides.html` if needed.

Observations:

1. The only way that I found to embed equations is by using the `:stem:` attribute, which automatically embeds a `<script>` tag pointing to the CDN version of MathJax. There is no way 
2. The `:revealjs_plugin_math: enabled` attribute is ignored, yet the documentation suggests that this should be the way to enable the official `math` Reveal.js plugin
[...]

PR #350, opened 2020-02-27
Fix for issue #348

akosma:
This PR provides a fix for #348, using a new attribute `mathjaxdir` set in the Asciidoc source.

Thanks to this, the user could for example `npm install mathjax@2.7.7` and then set the `mathjaxdir` attribute: `:mathjaxdir: node_modules/mathjax` to load the library locally instead of from the CDN.

If the attribute is not set, the behaviour is exactly the same as it is currently.

### DIFF
--- a/examples/mathjax.adoc
+++ b/examples/mathjax.adoc
@@ -1,0 +1,20 @@
+// .links
+// Demonstration of reveal.js and Mathjax integration
+// :include: //div[@class="slides"]
+// :header_footer:
+= MathJax
+:stem:
+:revealjsdir: https://cdn.jsdelivr.net/npm/reveal.js@3.9.2
+
+== Math Equation
+
+Using standard latexmath:[\LaTeX] syntax:
+
+[stem]
+++++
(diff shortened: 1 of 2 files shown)


akosma:
I've added the sample file, and wait for a decision regarding the naming of the attribute.

obilodeau:
Regarding `:mathjaxdir:` vs `:mathjax-dir:` I contacted Dan and Guillaume on Gitter too and there's no official decision at the moment. Let's proceed with `:mathjaxdir:`, we can always add an alias later if required.

examples/mathjax.adoc
```diff
@@ -0,0 +1,20 @@
+// .links
+// Demonstration of reveal.js and Mathjax integration
```
obilodeau:
```suggestion
// .mathjax
// Demonstration of our Mathjax integration
```

Mogztter:
> Regarding :mathjaxdir: vs :mathjax-dir: I contacted Dan and Guillaume on Gitter too and there's no official decision at the moment. Let's proceed with :mathjaxdir:, we can always add an alias later if required.

:+1:

PR #354, opened 2020-03-16
Added text-alignment to our column CSS layout

obilodeau:
Needed this for a deck. Adding it upstream now and followed bulma's naming (which I didn't use at first).


obilodeau:
Oups forgot to include screenshots I made.

Before:
adrjs-grid-text-align-before

After:
adrjs-grid-text-align-after

PR #356, opened 2020-03-17
background-video doesn't honor imagesdir attribute like it should

obilodeau:
Investigating what I thought was a problem with Asciidoctor, I realize that we are badly loading background-video local files.

Asciidoctor uses the [`media_uri` method](https://github.com/asciidoctor/asciidoctor/blob/bb47c333df0aaab6daefa1f4acd3ba3d2623c0c8/lib/asciidoctor/abstract_node.rb#L325) to lookup video files and we used it in the video template but for `background-video` this wasn't used.

This PR fixes that behavior.

Now, is this a backward-incompatible fix? It could break existing decks that relied on the wrong behavior. However, it was obviously broken and didn't follow what an Asciidoctor/AsciiDoc user would expect. I'm open to make this change part of a feature release but I would find it intense to bump a major version because of this. Guidance needed here.

Note
* Title slide `background-video` is already correctly implemented
* ~The first commit just exhibit the bug (I want to see it in travis), I have a local fix read to be pushed~ fix pushed

### DIFF
--- a/examples/video.adoc
+++ b/examples/video.adoc
@@ -34,7 +34,6 @@ video::https://s3.amazonaws.com/static.slid.es/site/homepage/v1/homepage-video-e
 [background-video="https://s3.amazonaws.com/static.slid.es/site/homepage/v1/homepage-video-editor.mp4,https://s3.amazonaws.com/static.slid.es/site/homepage/v1/homepage-video-editor.webm",options="loop,muted,notitle"]
 == background video at URL with options
 
-// FIXME doesn't load
 // Download this file and put in examples/images/ for this to be visible
 // https://github.com/obilodeau/asciidoctor-assets/blob/master/videos/synthwave.mp4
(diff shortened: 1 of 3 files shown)


Mogztter:
> Now, is this a backward-incompatible fix? It could break existing decks that relied on the wrong behavior. However, it was obviously broken and didn't follow what an Asciidoctor/AsciiDoc user would expect. I'm open to make this change part of a feature release but I would find it intense to bump a major version because of this. Guidance needed here.

It reminds me of: https://xkcd.com/1172/ (to a certain extend)
The behavior was defined by Asciidoctor and we should resolve images, audios and videos from the `imagesdir` directory, so we are fixing a bug.

Having said that you should put emphasis on the release note but I don't think we should do a major release. Otherwise, potentially every changes/bug fixes could potential break an existing workflow and thus requiring a major release.

Issue #316, opened 2020-01-22
background videos don't autostart in Chrome/Chromium: play request was interrupted


obilodeau:
Exact error message in the Chrome console:

```
Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause(). https://goo.gl/LdLk22
```

I tracked this down to a reveal.js behavior combined with a Chrome media playback API change and the fact that this project always sets an `id` on slides.

This works in a reveal.js html presentation:

```html
<div class="slides">
    <section>Empty slide to force interaction</section>
    <section data-background-video="https://s3.amazonaws.com/static.slid.es/site/homepage/v1/homepage-video-editor.mp4"></section>
</div>
```

But this doesn't:

```html
<div class="slides">
    <section>Empty slide</section>
    <section id="_customId" data-background-video="https://s3.amazonaws.com/static.slid.es/site/homepage/v1/homepage-video-editor.mp4"></section>
</div>
```

It affects reveal.js 3.7.0 but it could affect earlier versions. Version 3.8 (unsupported for now) is not affected.

Work-around is to use Firefox. I won't report upstream and I'll close this ticket. This is mostly meant as an FYI since I painfully tracked it down.

PR #357, opened 2020-03-17
Fixed embedded autoplay videos on Youtube and Vimeo

obilodeau:
iframes requires `allow="autoplay"` HTML attribute to allow content to autoplay now.

It might be a little more complicated since Chrome uses many heuristics to determine if a given domain can or can't autoplay iframes. See  https://developers.google.com/web/updates/2017/09/autoplay-policy-changes#iframe

My experience: using `file:///` URIs to load slidedecks I was never able to get block or inline videos hosted on youtube or vimeo to work reliably in the last few months. This was on both Chromium and Firefox. 

Also, in the meantime, we released 4.0 and `background-video` with autoplay works perfectly (see #316).

So coming back to this problem I diff'ed the iframes from both approaches (one done by reveal.js and the other by us) I found about that `allow="autoplay"` attribute. Adding it to the iframe in the video template fixes the problem.

### DIFF
--- a/templates/video.html.slim
+++ b/templates/video.html.slim
@@ -15,11 +15,12 @@
     - delimiter = '?'
     - loop_param = (option? 'loop') ? "#{delimiter}loop=1" : nil
     - src = %(#{asset_uri_scheme}//player.vimeo.com/video/#{attr :target}#{start_anchor}#{loop_param})
+    / We need to delegate autoplay into the iframe starting with Chrome 62 (and other browsers too)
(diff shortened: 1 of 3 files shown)


templates/video.html.slim
```diff
@@ -28,10 +29,14 @@
     - params << "end=#{attr :end}" if attr? :end
     - params << "loop=1" if option? 'loop'
     - params << "controls=0" if option? 'nocontrols'
+    /- params << "autoplay=1" if option? 'autoplay'
```
obilodeau:
```suggestion
```

PR #359, opened 2020-03-17
Added example loading Mathjax from CDN

akosma:
This adds an example file showing how to specify the `:mathjaxdir:` attribute, which works with a URL or a local path.


obilodeau:
I guess we had a misunderstanding, I wanted a case that would not render correctly without a newer version of MathJax but that will do. Thanks!

PR #361, opened 2020-03-24
Updated Mathjax to latest version available (fixes #355)

### LINKED ISSUE #355: Should we bump to latest MathJax 2.7.x?
Request for comments

MathJax 2.4 is about to be 6 years old. The latest stable release 2.7.5 is 2 years old. It seems pretty stable. Trying the example that @akosma provided with it worked.

Release notes of in-between versions don't reveal any backward-incompatible changes:
* https://www.mathjax.org/mathjax-v2-5-now-available/
* https://www.mathjax.org/mathjax-v2-6-now-available/
* https://www.mathjax.org/mathjax-v2-7-now-available/
* Several 2.7.x point releases with the latest being: https://www.mathjax.org/mathjax-v2-7-5-now-available/

I think we should update our default version to 2.7.5 given that it seems backward-compatible, any objections?

obilodeau:


obilodeau:
I tried 2.7.8, it doesn't work because cloudflare is not hosting it (see #355), it would require breaking out of `cdn_base` and I'm not willing to do this in a stable release given that revealjs 4.0 is out now.

PR #368, opened 2020-05-26
resolves #367 allow to configure autoslide timer per slide

### LINKED ISSUE #367: Reveal JS autoslide problem
I want to use Reveal JS autoslide for a presentation, and would overwrite this function with `:data-autoslide:` integer like it is descript in this [document](https://asciidoctor.cn/docs/asciidoctor-revealjs/#reveal-js-optionst) but it won't change his first value.

```
asciidoctor-revealjs -v 
Asciidoctor reveal.js 4.0.1 using Asciidoctor 2.0.10 [https://asciidoctor.org]
Runtime Environment (ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux-gnu]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)
```

ggrossetie:
resolves #367

### DIFF
--- a/examples/autoslide.adoc
+++ b/examples/autoslide.adoc
@@ -1,0 +1,22 @@
+= Autoslide!
+// automatically proceed to the next slide after 5 seconds
+:revealjs_autoSlide: 5000
+
+== Slide 0
+
+I can do it in 5 seconds!
+
+// on this slide we need 10 seconds
+[autoslide=10000]
+== Slide 1
+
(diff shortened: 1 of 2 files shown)


obilodeau:
LGTM, thanks! Especially for taking the time to help out like you did in #367 *and* do the fix yourself in the end. Teaching our users to contribute I think will pay off in the long term but it doesn't work every time.

PR #383, opened 2020-11-05
resolves #241 set data- attributes

### LINKED ISSUE #241: settomg data-fragment-index on elements
It would be great to have possibility to set data-fragment-index on an element. This could allow to make simultaneous appearence or disappearence or whatever else of several elements.

This issue https://github.com/asciidoctor/asciidoctor/issues/1305 would solve the  problem, but it is not implemented yet.

General solution could look like:

[.fragment.highlight-red, data-fragment-index=1]#text1#
[.fragment.highlight-red, data-fragment-index=0]#text2#
[.fragment.highlight-red, data-fragment-index=1]#text3#

On the 1st step we get text2 turning red, on the second -- text1 and text3 turning red.

Mogztter:
Please note that for now this feature is only available on inline images and paragraphs.
I want to make sure that we are on the same page before implementing this feature on other nodes.

I think it can makes sense to support this feature on:

- [x] admonition
- [x] audio (why not, people might find a creative usage)
- [x] colist
- [x] dlist
- [x] example
- [x] image
- [x] inline_anchor
- [x] inline_image
- [x] listing
- [x] literal
- [x] olist
- [x] open
- [x] paragraph
- [x] quote
- [x] stem
- [x] table
- [x] ulist
- [x] verse
- [x] video
- [x] sidebar
- [x] inline_quoted → :information_source: roles and attributes must be defined using the API
- [x] inline_button → :information_source: roles and attributes must be defined using the API
- [x] inline_footnote → :information_source: roles and attributes must be defined using the API
- [x] inline_kbd → :information_source: roles and attributes must be defined using the API
- [x] inline_menu → :information_source: roles and attributes must be defined using the API

It's worth noting that Asciidoctor does not support attributes on inline marks/quoted/phrase. For instance:

```
[.fragment.highlight-red,data-fragment-index=1]#text1#
```

In the above definition, `data-fragment-index=1` will be ignored (i.e. this attribute will not be available on the inline node).

resolves #241


Mogztter:
fragment-revealjs

Mogztter:
@obilodeau What do you think?

Mogztter:
@obilodeau Any opinion on it? Should I proceed and implement this feature on other node types?

obilodeau:
> @obilodeau Any opinion on it? Should I proceed and implement this feature on other node types?

It looks great. Go ahead! Your example seems quite extensive as well. Good job.

Mogztter:
> It looks great. Go ahead! Your example seems quite extensive as well. Good job.

Thanks!
Yes I'm adding an example/test for each node type :smile:

PR #389, opened 2020-11-28
resolves #385 migrate the documentation to Antora

### LINKED ISSUE #385: Migrate docs to Antora
We should use a concise README with a link to the complete documentation.
The documentation will be available at https://docs.asciidoctor.org (upcoming)

Mogztter:
resolves #385


Mogztter:
You can take a look at the documentation (**preview**): https://deploy-preview-24--asciidoctor-docs.netlify.app/asciidoctor-reveal.js/current/

obilodeau:
It is looking good! Just a quick note to say that links using `{uri-revealjs-gh}` attribute don't work.

Mogztter:
@obilodeau thanks for your review, I will fix that :ok_hand:

docs/antora.yml
```diff
@@ -0,0 +1,18 @@
...
+title: Asciidoctor reveal.js
+version: '4.0'
+nav:
+  - modules/setup/nav.adoc
+  - modules/converter/nav.adoc
+  - modules/ROOT/nav.adoc
+  - modules/project/nav.adoc
+asciidoc:
+  attributes:
+    uri-project-repo: 'https://github.com/asciidoctor/asciidoctor-reveal.js'
```
mojavelinux, line 11: Please use the url- prefix for URLs instead of uri-. URL is now the preferred over URI as specified by the What WG. See https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#URLs_and_URNs

docs/modules/ROOT/nav.adoc
```diff
@@ -0,0 +1 @@
+* xref:showcase.adoc[Showcase]
```
mojavelinux, line 1: It's better to leave the text blank here and either allow Antora to use the doctitle or set the navtitle attribute in the document header.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -0,0 +1,13 @@
+= Reveal.js converter for Asciidoctor
```
mojavelinux, line 1: The reveal.js project seems very deliberate about keeping the "r" lowercase. So I think this title should read "reveal.js".

Also, define the `navtitle` attribute on this page as follows.

```
:navtitle: Introduction
```

This will allow the UI to configure the breadcrumbs correctly for this page.

Mogztter: You're right but the capitalization feels a bit weird when other words are capitalized:

#### reveal.js Plugins

#### reveal.js Converter for Asciidoctor

Mogztter: @mojavelinux I've added `:navtitle: Introduction` but the breadcrumb stills show "reveal.js Converter for Asciidoctor": 

[`0897e6b` (#389)](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/389/commits/0897e6b7b49b99b2d619383169532613421db996)


breadcrumb

Did I miss something?

mojavelinux: Yes. Here's the logic: https://github.com/asciidoctor/asciidoctor-docs-ui/blob/main/src/partials/breadcrumbs.hbs#L11

mojavelinux: > the capitalization feels a bit weird when other words are capitalized

It may seem weird, but reveal.js goes out of its way to not capitalize the project name on its site, so we should respect that. According to the English style guides, a proper noun which is not capitalized should never be capitalized.

Mogztter: So it should be "Asciidoctor reveal.js Converter Documentation". Maybe we can remove the "Asciidoctor", what do you think?

mojavelinux: That's the correct title. You could also drop "Converter". The important part is to set the `navtitle` because I will update the UI to use it. Then, the doctitle can be whatever you want it to be. Though I think "Asciidoctor reveal.js Documentation" seems clear enough.

Mogztter: Done!

ok

Mogztter:
@mojavelinux Thanks for your review, if you have nothing left I think we can merge?

mojavelinux:
What's your position on the naming of the component? Should it be asciidoctor-reveal.js or reveal.js-converter? See https://github.com/asciidoctor/asciidoctor.org/issues/977

mojavelinux:
Btw, the Maven plugin is using the component name maven-plugin.

Mogztter:
`reveal.js-converter` sounds good, I will rename

Mogztter:
@mojavelinux It's a chicken and egg situation, if I merge the branch on master, the README will point to a not yet existing documentation site: https://asciidoctor-docs.netlify.app/reveal.js-converter/current/

I will revert the changes on the README so I can merge this pull request on master and do a follow-up pull request to update the README once the documentation site is live. What do you think?

mojavelinux:
@Mogztter I think that's a good plan. The URL isn't right anyway since that URL will go away once we make docs.asciidoctor.org live (technically it will redirect, but it's still an internal URL).

Issue #304, opened 2020-01-11
Missing brand icons in v3.0


gquintana:
Since I upgraded to ascidoctor-reveal.js from v2.0.1 (Font Awesome 4.7) to v3.0.0 (Font Awesome 5). I miss social network icons. All other icons are alright.

For example, the following is not rendered properly:
```
icon:twitter[] twitter
icon:github[] github
```
![Firefox_Screenshot_2020-01-11T16-32-24 051Z](https://user-images.githubusercontent.com/755587/72207332-7de42b00-3498-11ea-9085-d53ab2ad4ea3.png)
This kind of icon should be rendered as
```
<i class="fab fa-twitter"></i>
```
Not
```
<i class="fa fa-twitter"></i>
```
How can I do it?

obilodeau:
I realize now that the problem is that we updated before asciidoctor started supporting it (see https://github.com/asciidoctor/asciidoctor/issues/2535). It's a pity that Font Awesome did a pretty breaking change.

I'll try the shim offered by Font Awesome and if that doesn't work I'll switch back to the latest Font Awesome 4

obilodeau:
I can confirm that adding:

```html
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.8.2/css/v4-shims.min.css">
```

after

```html
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.8.2/css/all.min.css">
```

in the generated HTML file fixes the issue for twitter (and most likely others).

I'll do a PR soon and probably a release. In the meantime you can add it manually after the conversion.

I just read the Font-Awesome changelog and I'll update to 5.12.0 at the same time since there's nothing breaking and a bunch of new icons.

Mogztter:
Currently, it's not possible because the template is always using `fa` as a CSS class:

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/95fa8edd770f1733b9c9c02ce4ea64b3aab403df/templates/inline_image.html.slim#L3


In [Asciidoctor Web PDF](https://github.com/Mogztter/asciidoctor-pdf.js/blob/ee8efaee4df18be42485b4e3eb6eadc4ebffe0b6/lib/document/templates.js#L192), I've added an attribute named `set` to support multiple styles or icons libraries:

```
icon:address-book[] icon:address-book[set=far]
```

[...]

PR #392, opened 2020-12-01
Upgrade Font Awesome to 5.15.1

### LINKED ISSUE #363: Upgrade font-awesome to 5.13.0
I'm sure some of us could use the covid-19 fonts from 5.13.0 

https://fontawesome.com/changelog/latest

Mogztter:
Introduce an attribute to configure the version (`font-awesome-version`).

I also found a workaround to support brand icons.
You can declare the following:

```
icon:rust[2x,role=fab]
```

And then, add the following styles in your custom stylesheet:

```css
/* Font Awesome 5.x Brand compatibility */
span.icon.fab > i.fa {
  font-family: "Font Awesome 5 Brands";
  font-weight: initial;
}
```

resolves #363


mojavelinux:
> icon:rust[2x,role=fab]

This is likely out of scope for this change, but I at least want to bring up the point. I think we should standardize on `set` like is used by Asciidoctor PDF. See https://github.com/asciidoctor/asciidoctor-pdf#font-based-icons

Mogztter:
> This is likely out of scope for this change, but I at least want to bring up the point

Yes, I was just experimenting :test_tube: 
It's worth noting that I didn't introduce any change to the converter, I was showcasing a workaround that can be used today.

> I think we should standardize on set like is used by Asciidoctor PDF. See asciidoctor/asciidoctor-pdf#font-based-icons

I was trying to address Olivier concern about introducing a new syntax: https://github.com/asciidoctor/asciidoctor-reveal.js/issues/304#issuecomment-573485940 but I also think that we should use the `set` attribute.
Anyway, I will create a dedicate issue and pull request to track this change.

mojavelinux:
Super!

Btw, if you want to implement the workaround, you might consider defining both role and set and eventually dropping role once support for set is implemented. Just an idea.

PR #416, opened 2021-02-12
Fix subscripts erroneously mapped to superscripts

### LINKED ISSUE #415: Subscripts are outputted as superscripts
The bug seems to be introduced in a3a13667222839e784fbdde8988e1833df951d05.

Outputs are fine if using the HTML backend.

9ao9ai9ar:
Fixes #415


Mogztter:
Thanks @9ao9ai9ar 
I will add a test to avoid future regressions :sweat_smile:

Issue #419, opened 2021-02-26
Favicon is not included in HTML


bbenno:
## Expected
Using the document attribute [`:favicon:`](https://docs.asciidoctor.org/asciidoctor/latest/html-backend/favicon/) results in adding
```html
<link rel="icon" type="image/x-icon" href="favicon.ico">
```
to HTML header as in [Asciidoctor](https://github.com/asciidoctor/asciidoctor) using HTML backend.

## Actual
No such HTML element is inserted.

Mogztter:
Hey @bbenno!
That would be a great addition, do you want to submit a pull request? You can take a look at the original implementation in Asciidoctor:

https://github.com/asciidoctor/asciidoctor/blob/3c7d2e8255d2f75dede0aab42ca32948ca4360e4/lib/asciidoctor/converter/html5.rb#L113-L123

You can add the logic just after this line: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/7bbad8d4b7e21c5beb8bce17eb12532d30c49e62/templates/document.html.slim#L18

We are using [Slim templates](http://slim-lang.com/). Let me know if you need help to get started.

bbenno:
Hey @Mogztter, thx for the quick reply an the references.
Please have a look at the PR.

PR #420, opened 2021-02-26
Add favicon to HTML if its attribute is present in AsciiDoc

bbenno:
Source of added code can be found in asciidoctor repository
https://github.com/asciidoctor/asciidoctor/blob/3c7d2e8255d2f75dede0aab42ca32948ca4360e4/lib/asciidoctor/converter/html5.rb#L113-L123

rel #419

### DIFF
--- a/examples/favicon.adoc
+++ b/examples/favicon.adoc
@@ -1,0 +1,10 @@
+// .favicon
+// Include favicon
+// :include: //head/link[@rel="icon"]
+// :header_footer:
+= Favicon
+:favicon:
+
+== First Step
+
+Check my favicon!
--- a/examples/favicon.adoc
+++ b/examples/favicon.adoc
@@ -1,10 +1,13 @@
 // .favicon
-// Include favicon
+// Include default favicon (`./favicon.ico`)
 // :include: //head/link[@rel="icon"]
 // :header_footer:
 = Favicon
 :favicon:
 
-== First Step
-
-Check my favicon!
+// .favicon-by-path
+// Include favicon by path
+// :include: //head/link[@rel="icon"]
+// :header_footer:
+= Favicon by Path
+:favicon: img/favicon.png
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -16,6 +16,15 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     - [:description, :keywords, :author, :copyright].each do |key|
       - if attr? key
         meta name=key content=(attr key)
+    - if attr? 'favicon'
+      - if (icon_href = attr 'favicon').empty?
+        - icon_href = 'favicon.ico'
+        - icon_type = 'image/x-icon'
+      - elsif (icon_ext = File.extname icon_href)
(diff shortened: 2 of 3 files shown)


Mogztter:
It looks good!
Would you mind adding a test?
You can create a file named `favicon.adoc` in `examples` directory:

```
// .favicon
// :include: //head/link[@rel="icon"]
// :header_footer:
= Favicon
:favicon:

== First Step

Check my favicon!
```

And then add `favicon.html` in `test/doctest`:

```html
<!-- .favicon -->
<link rel="icon" type="image/x-icon" href="favicon.ico">
```

It should do the trick!
Basically, you can create as many examples as you want and then you need to assert the HTML produced in `test/doctest`.

Some details about the comments in `favicon.adoc`.
Here you define a unique id:

```
// .favicon
```

Please note that you will need to use the same id in the HTML (`<!-- .favicon -->`)

This is a xpath to check only matching nodes (and not the complete HTML):

```
// :include: //head/link[@rel="icon"]
```

And last, we want to generate a full document with header and footer (i.e, including `<head>`):

```
:header_footer:
```

bbenno:
> Would you mind adding a test?

Would be a pleasure! Basically I did not figure out the process of testing yesterday. So, thank you very much for the great support!

bbenno:
@Mogztter I would like to add another test for the usage of a favicon with custom file path.
```adoc
// .custom-favicon
// :include: //head/link[@rel="icon"]
// :header_footer:
= Custom Favicon
:favicon: img/favicon.png
```
This should result in:
```html
<!-- .custom-favicon -->
<link rel="icon" type="image/png" href="img/favicon.png">
```

Is there a way to do this within the same test file `examples/favicon.adoc`?
As far as I can see this is not possible, isn't it?
If so, I would like to add the above test in `examples/favicon-custom.adoc`...

Mogztter:
> Would be a pleasure! Basically I did not figure out the process of testing yesterday. So, thank you very much for the great support!

You're welcome!
We should probably add these instructions in https://docs.asciidoctor.org/reveal.js-converter/latest/project/hacking/ (https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/docs/modules/project/pages/hacking.adoc)

> I would like to add another test for the usage of a favicon with custom file path.

I think it's possible, `asciidoctor-doctest` (the underlying library) is using this document: `https://github.com/asciidoctor/asciidoctor-doctest/blob/master/data/examples/asciidoc/document.adoc`

Did you try the following?


```
// .favicon
// :include: //head/link[@rel="icon"]
// :header_footer:
= Favicon
:favicon:

// .custom-favicon
// :include: //head/link[@rel="icon"]
// :header_footer:
= Custom Favicon
:favicon: img/favicon.png
```

> Is there a way to do this within the same test file examples/favicon.adoc?
As far as I can see this is not possible, isn't it?
If so, I would like to add the above test in examples/favicon-custom.adoc

Indeed, if it does not work you can create `favicon-custom.adoc` (I'm fine with that).

Mogztter:
LGTM, thanks @bbenno 👍

PR #421, opened 2021-02-27
Drop thread_safe and concurrent-ruby dependencies

slonopotamus:
Seems like they serve no purpose in asciidoctor-reveal.js and most likely are artifacts from ancient Ruby versions

### DIFF
--- a/asciidoctor-revealjs.gemspec
+++ b/asciidoctor-revealjs.gemspec
@@ -31,8 +31,6 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
 
   s.add_runtime_dependency 'asciidoctor', ['>= 2.0.0', '< 3.0.0']
-  s.add_runtime_dependency 'thread_safe', '~> 0.3.5'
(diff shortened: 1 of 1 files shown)


Mogztter:
> Seems like they serve no purpose in asciidoctor-reveal.js and most likely are artifacts from ancient Ruby versions

I think you are right!
For reference, [Dan said](https://github.com/asciidoctor/asciidoctor-reveal.js/pull/216#issuecomment-434186712):

> For compatibility with 1.5.8, you should add a dependency on the concurrent-ruby gem. You can still depend on thread_safe too. As long as both gems are present, it will work on both 1.5.7 and 1.5.8 without any warnings.

> Asciidoctor core has a policy of having 0 dependencies. On top of that, it's only a recommended dependency, not a required one.

(^^^ speaking about thread_safe dependency)

> And, btw, it's now concurrent-ruby, not thread_safe. Though, for it to work without warning on both 1.5.7 and 1.5.8, you'll need both on the load path.

Mogztter:
Thanks, less dependencies is 💚

Issue #226, opened 2018-11-18
background image in CSS not possible anymore


bcouetil:
First, I want to thank you very much for the good work on Asciidoctor and Asciidoctor-reveal. Using it extensively for both documentation and presentations. Loving it, loving you guys :)

I was on 1.1.3 version and wanted to use the master branch to be able to use plugins. But now the rendering behave deferently and I did not find any way around.

Here is my slides in 1.1.3 : https://bcouetil.gitlab.io/academy/PRES-asciidoc-dark.html

You added this from #16 in HTML  : 
`  if (slide.getAttribute('data-background-color')) return;
  var bgColor = getComputedStyle(slide).backgroundColor;
  if (bgColor !== 'rgba(0, 0, 0, 0)' && bgColor !== 'transparent') {
    slide.setAttribute('data-background-color', bgColor);
    slide.style.backgroundColor = 'transparent';
  }})`

My background image is in CSS, with nothing about style in the .adoc. Now, with new version, the body background is forced to be the same as the slides background.

What do you suggest ?

obilodeau:
What CSS are you using to set the background image? It looks like it's done with a custom reveal.js theme, is there a place I can look at the source?

Once I know how it's set, I think we can query for it in Javascript and do the background color assignment only if there's no background defined or something like that.

p.s.: Thanks for the kind words :smile:

bcouetil:
Thank you for the quick answer.

Everything is in reveal-zenika-dark.css, viewable in code inspection.

For the global background :
`body {
    background: no-repeat center url(./logo-zenika.png);
    background-size: 60%;
    background-color: black;
}`

For the slide background : 
`.reveal .slides section {
    background: #222;
}`

mojavelinux:
> Loving it, loving you guys :)

:heart_eyes:

mojavelinux:
I had not considered the case where the canvas has a different background color from the slides (since that's not how I make my slides).

In order to support this, I think we'd need to designate a role to control whether the color from the slide gets propagated to the canvas. One possibility is to designate the "layer" role as a way to disable this behavior. Another possibility, which would be more compatible with existing presentations, is to disable this behavior by default and only activate it if the "canvas" role is present. In other words:

```
[.red.canvas]
== Slide Title
```

Then the logic in the JavaScript would be:

```
if (!slide.classList.contains('canvas')) return;
```

[...]

PR #427, opened 2021-06-06
resolves #370 add support for reveal.js 4.x

### LINKED ISSUE #370: reveal.js 4.0 support
4.0.0 was [released](https://github.com/hakimel/reveal.js/releases/tag/4.0.0) last week and with it enough breaking changes for us to mandate another breaking one.

However, one thing we will need to do in the meantime is to change all of our links to the reveal.js README to point to their new 3.9 branch. They released a dedicated documentation site and stripped their README of all the stuff. I think master and maybe -1, -2 major releases should point to the right location (given it's only some vim regex wizardry at this point) and adjust all the links for our next major release.

edl7878:
Reveal.js 4.x assets path is varied from 3.9.x, in this pull request,  asciidoctor-reveal.js code is modified:

* paths are directed to the right place
* since print related CSSes are included in reveal.css in reveal.js 4.x, print css injections are moved out in code base


Mogztter:
Thanks @edl7878

Since you've made some changes on the templates, you will need to update the tests as well.

Mogztter:
Could you please also update `revea.js` version in: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/e89b5af7041ceb02de5bb35810fb0f563b97a8e6/package.json#L52

Mogztter:
@edl7878 you can run tests locally using:

```
bundle exec rake build
bundle exec rake test
```

You will need to update the HTML files in https://github.com/asciidoctor/asciidoctor-reveal.js/tree/master/test/doctest 

For instance, in `test/doctest/custom-layout.html`, you will need to fix two lines:

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/e89b5af7041ceb02de5bb35810fb0f563b97a8e6/test/doctest/custom-layout.html#L10

It should be `<script src="reveal.js/dist/reveal.js"></script>`

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/e89b5af7041ceb02de5bb35810fb0f563b97a8e6/test/doctest/custom-layout.html#L170

It should be `{ src: 'reveal.js/plugin/zoom/zoom.js', async: true },`

And here's the relevant test failure (taken from GitHub Action log):

```
(See full trace by running task with --trace)
✗  Failure: custom-layout:custom-layout
   The three different ways to hide slide titles.

          </div>
        </section>
      </div>
   E  <script src="reveal.js/js/reveal.js"></script><script>
   A  <script src="reveal.js/dist/reveal.js"></script><script>
        Array.prototype.slice.call(document.querySelectorAll('.slides section')).forEach(function(slide) {
          if (slide.getAttribute('data-background-color')) return;
          // user needs to explicitly say he wants CSS color to override otherwise we might break custom css or theme (#226)
        
          // Optional libraries used to extend on reveal.js
          dependencies: [
   E          { src: 'reveal.js/plugin/zoom-js/zoom.js', async: true },
   A          { src: 'reveal.js/plugin/zoom/zoom.js', async: true },
              { src: 'reveal.js/plugin/notes/notes.js', async: true }
          ],
        
```

edl7878:
> @edl7878 you can run tests locally using:
> 
> ```
> bundle exec rake build
> bundle exec rake test
> ```
> 
> You will need to update the HTML files in https://github.com/asciidoctor/asciidoctor-reveal.js/tree/master/test/doctest
> 
> For instance, in `test/doctest/custom-layout.html`, you will need to fix two lines:
> 
> https://github.com/asciidoctor/asciidoctor-reveal.js/blob/e89b5af7041ceb02de5bb35810fb0f563b97a8e6/test/doctest/custom-layout.html#L10
> 
> It should be `<script src="reveal.js/dist/reveal.js"></script>`
> 
> https://github.com/asciidoctor/asciidoctor-reveal.js/blob/e89b5af7041ceb02de5bb35810fb0f563b97a8e6/test/doctest/custom-layout.html#L170
> 
> It should be `{ src: 'reveal.js/plugin/zoom/zoom.js', async: true },`
> 
> And here's the relevant test failure (taken from GitHub Action log):
> 
> ```
> (See full trace by running task with --trace)
> ✗  Failure: custom-layout:custom-layout
>    The three different ways to hide slide titles.
> 
>           </div>
>         </section>
>       </div>
>    E  <script src="reveal.js/js/reveal.js"></script><script>
>    A  <script src="reveal.js/dist/reveal.js"></script><script>
>         Array.prototype.slice.call(document.querySelectorAll('.slides section')).forEach(function(slide) {
>           if (slide.getAttribute('data-background-color')) return;
>           // user needs to explicitly say he wants CSS color to override otherwise we might break custom css or theme (#226)
>         
>           // Optional libraries used to extend on reveal.js
>           dependencies: [
>    E          { src: 'reveal.js/plugin/zoom-js/zoom.js', async: true },
>    A          { src: 'reveal.js/plugin/zoom/zoom.js', async: true },
>               { src: 'reveal.js/plugin/notes/notes.js', async: true }
>           ],
>         
> ```

There are 15 file issues, are these files in the test manually edited?  If I modify them directly, could it be substituted by another pass of code generation?

edl7878:
test file modified. Still has 2 errors, but I think the modification is correct:

···
   E  <link href="https://cdn.jsdelivr.net/npm/reveal.js@4.1.1/dist/reset.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@4.1.1/dist/reveal.css" rel="styleshee   A  <link href="https://cdn.jsdelivr.net/npm/reveal.js@3.9.2/dist/reset.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@3.9.2/dist/reveal.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@3.9.2/dist/theme/black.css" id="theme" rel="stylesheet">
      <div class="slides">
        <section class="title" data-state="title">
          <h1>MathJax</h1>


✗  Failure: revealjs-custom-theme:revealjs-custom-theme
   please note that in this specific case it would be better to use :revealjs_theme: beige.

   E  <link href="reveal.js/dist/reset.css" rel="stylesheet"><link href="reveal.js/dist/reveal.css" rel="stylesheet"><link href="reveal.js/dist/theme/beige.css" id="theme" rel="stylesheet">
   A  <link href="reveal.js/dist/reset.css" rel="stylesheet"><link href="reveal.js/dist/reveal.css" rel="stylesheet"><link href="reveal.js/css/theme/beige.css" id="theme" rel="stylesheet">
···

Mogztter:
`mathjax-cdn.adoc` is using a custom version of reveal.js:

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/e89b5af7041ceb02de5bb35810fb0f563b97a8e6/examples/mathjax-cdn.adoc#L7

So the expectation is that the specified version of reveal.js will be used.
I think we should update the value of the `revealjsdir` attribute to use an older version of reveal.js 4.x `:revealjsdir: https://cdn.jsdelivr.net/npm/reveal.js@4.0.2`.

The same goes for `revealjs-custom-theme.adoc`.


Please note that reveal.js 4.1.3 was just released.

test/doctest/mathjax-cdn.html
```diff
@@ -1,5 +1,5 @@
 <!-- .mathjax-cdn -->
-<link href="https://cdn.jsdelivr.net/npm/reveal.js@3.9.2/css/reset.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@3.9.2/css/reveal.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@3.9.2/css/theme/black.css" id="theme" rel="stylesheet">
+<link href="https://cdn.jsdelivr.net/npm/reveal.js@4.1.1/dist/reset.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@4.1.1/dist/reveal.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@4.1.1/dist/theme/black.css" id="theme" rel="stylesheet">
```
Mogztter, line 2: We should update the attribute `:revealjsdir:` in `mathjax-cdn.adoc` and update the expectations:

```
:revealjsdir: https://cdn.jsdelivr.net/npm/reveal.js@4.0.2
```

```suggestion
<link href="https://cdn.jsdelivr.net/npm/reveal.js@4.0.2/dist/reset.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@4.0.2/dist/reveal.css" rel="stylesheet"><link href="https://cdn.jsdelivr.net/npm/reveal.js@4.0.2/dist/theme/black.css" id="theme" rel="stylesheet">
```

test/doctest/revealjs-custom-theme.html
```diff
@@ -1,2 +1,2 @@
 <!-- .revealjs-custom-theme -->
-<link href="reveal.js/css/reset.css" rel="stylesheet"><link href="reveal.js/css/reveal.css" rel="stylesheet"><link href="reveal.js/css/theme/beige.css" id="theme" rel="stylesheet">
+<link href="reveal.js/dist/reset.css" rel="stylesheet"><link href="reveal.js/dist/reveal.css" rel="stylesheet"><link href="reveal.js/dist/theme/beige.css" id="theme" rel="stylesheet">
```
Mogztter, line 2: We should update `revealjs-custom-theme.adoc`:

```diff
-:revealjs_customtheme: reveal.js/css/theme/beige.css
+:revealjs_customtheme: reveal.js/dist/theme/beige.css
```

edl7878:
Does npm has no reveal.js 4.1.1? And I will roll back several modifications? Or, 4.0.2 is preferred?

Mogztter:
> Does npm has no reveal.js 4.1.1? And I will roll back several modifications? Or, 4.0.2 is preferred?

Latest version available on npmjs is 4.1.3: https://www.npmjs.com/package/reveal.js
I was suggesting to define `:revealjsdir: https://cdn.jsdelivr.net/npm/reveal.js@4.0.2` to use a different version (for testing purpose).

Basically, we want to check that when `revealjsdir` is defined, the Asciidoctor reveal.js converter will use it.

edl7878:
OK, if needed, I will target to 4.1.3.

Mogztter:
We are almost good (according to CI), last error is:

```
npm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
```

Could you please run `npm install` locally and commit `package-lock.json` file?

Mogztter:
One more issue catch by the tests suite:

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/ba16105cedb80171a127d543215b53db2b3f01a4/test/smoke.js#L21

You need to update the assertion:

```js
expect(result).to.contain('<script src="node_modules/reveal.js/dist/reveal.js">') 
```

maxandersen:
how would i actually go about using this pr's build locally to render my own asciidoc reveal.js documents ? (i got the build working - just not sure how to refer to the local built for use with asciidcotor revealjs ?

maxandersen:
I figured it out.

curl -L -o m.zip https://github.com/hakimel/reveal.js/archive/master.zip

unzip it so its named reveal.js

then run `bundle exec asciidoctor-revealjs test.adoc`

reason im trying this is I really like the [auto-animate feature](https://revealjs.com/auto-animate/) in reveal.js 4.

I kinda got it working by adding:`
`data-auto-animate=(attr 'auto-animate')` in the templates but:

1) i have to use auto-animate=true in my adoc to have it pickedup - how do i just add the attribute if present ?
2) auto-animate relies on stable data-id attributes; any tips on how to ensure such gets applied to all kind of blocks? is it required to be manually added to every  template?
3) the animations on reveal.js are way smoother than what we get rendered - not sure if just a different default style?

Mogztter:
> i have to use auto-animate=true in my adoc to have it pickedup - how do i just add the attribute if present ?

You can use `bool_data_attr` method from the helper: 

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/ba16105cedb80171a127d543215b53db2b3f01a4/templates/helpers.rb#L35-L42

```slim
data-auto-animate=(bool_data_attr 'auto-animate')
```

> auto-animate relies on stable data-id attributes; any tips on how to ensure such gets applied to all kind of blocks? is it required to be manually added to every template?

You can already define data attributes on virtually any blocks/elements: https://asciidoctor-revealjs-examples.netlify.app/release-4.1.html#/_data_attributes

> the animations on reveal.js are way smoother than what we get rendered - not sure if just a different default style?

Not sure about this one... Apparently, it's possible to override animation settings: https://revealjs.com/auto-animate/#animation-settings

maxandersen:
> You can already define data attributes on virtually any blocks/elements: https://asciidoctor-revealjs-examples.netlify.app/release-4.1.html#/_data_attributes

cool - unfortunately seems it won't honor it on sections ? otherwise I should be able to just do:

```adoc
[data-auto-animate=0]
== Test 

[source,java,data-id=code]
----
System.out.println('What upo!');
----
```
but here only the data-id gets through.

maxandersen:
> ```
> data-auto-animate=(bool_data_attr 'auto-animate')
> ```

I tried this but it doesn't do what I tried to ask for :) - let me try again:

With my previous fix and your suggestion above I need to write like this:

```adoc
[auto-animate=true]
== Some section

[source,java,data-id="mycode"]
----
System.out.println('What upo!');
----
```

What I really want is:

```adoc
[auto-animate]
== Some section

[source,java,data-id="mycode"]
----
System.out.println('What upo!');
----
```

i.e. not having to add `=true` as in reveal.js they just require the presence of `auto-animate` the value does not seem to matter.

Mogztter:
In this case we should use an option and not an attribute. In the following example `auto-animate` is actually a block name:

```
[auto-animate]
== Some section
```

The reason is that we cannot disambiguate a block name from an attribute without a value. I believe that this is one of the reason why we have option:

```
[%auto-animate]
== Some section
```

> cool - unfortunately seems it won't honor it on sections ? otherwise I should be able to just do:



Indeed, on `<section>` data attributes are explicit:

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/ba16105cedb80171a127d543215b53db2b3f01a4/templates/section.html.slim#L50-L66


I would recommend the following:

```slim
data-auto-animate=((attr? 'auto-animate') || (option? 'auto-animate'))
```

I haven't tested it but in theory the following should work:


```
[%auto-animate]
== Some section
```

```
[auto-animate=]
== Some section
```

edl7878:
> In this case we should use an option and not an attribute. In the following example `auto-animate` is actually a block name:
> 
> ```
> [auto-animate]
> == Some section
> ```
> 
> The reason is that we cannot disambiguate a block name from an attribute without a value. I believe that this is one of the reason why we have option:
> 
> ```
> [%auto-animate]
> == Some section
> ```
> 
> > cool - unfortunately seems it won't honor it on sections ? otherwise I should be able to just do:
> 
> Indeed, on `<section>` data attributes are explicit:
> 
> https://github.com/asciidoctor/asciidoctor-reveal.js/blob/ba16105cedb80171a127d543215b53db2b3f01a4/templates/section.html.slim#L50-L66
> 
> I would recommend the following:
> 
> ```
> data-auto-animate=((attr? 'auto-animate') || (option? 'auto-animate'))
> ```
> 
> I haven't tested it but in theory the following should work:
> 
> ```
> [%auto-animate]
> == Some section
> ```
> 
> ```
> [auto-animate=]
> == Some section
> ```

```
[data-auto-animate=true]
## test adoc1
```
won't work. data-auto-animate can't map to section tag, while common block is ok, likes this:

```
## test adoc1

for test2

[data-auto-animate=true]
for test1
```

edl7878:
How to reopen this pull request? It seems I closed it accidentally.

edl7878:
I think the translating layer, makes things harder for users, such as:

```
'fragment' if (option? :step) || (attr? 'step')
```

```
data-background-iframe=(attr "background-iframe")
```

why not just use reveal.js original keywords, which is easier for users.

except when use options, they have to use %, if they want cleaner output.

edl7878:
For now, data-auto-animate is workable. test file:

```
# test
:revealjsdir: reveal.js

[%auto-animate]
## test adoc

[.r-fit-text]
for test1

[%auto-animate]
## test adoc1

for test2

[data-auto-animate=true]
for test1
```

All settings are added to section element, except `data-auto-animate-delay`, which is not applicable for section.

Not fully test all settings, for now.

Mogztter:
> I think the translating layer, makes things harder for users, such as:
'fragment' if (option? :step) || (attr? 'step')

I think the idea behind this abstraction is to make your content "portable" (or at least not tightly coupled to a specific backend).
As far as I remember, you can use the role `fragment` but it might not be supported on all elements.

```
[.fragment]
This is a paragraph.
```

> data-background-iframe=(attr "background-iframe")
why not just use reveal.js original keywords, which is easier for users.

In this case we are using the same name (without the `data-` prefix which is an implementation detail).
I guess we could support both name, but in my opinion `background-iframe` should be preferred:

```
[data-background-iframe=https://slides.com]
== Section
```

```
[background-iframe=https://slides.com]
== Section
```

edl7878:
Yes, I agree with you that after getting rid of `data-` prefix, it is more independent from reveal.js.

I just think that from the user's point of view, when they check the documentation of reveal.js, and find out there is a class could introduce a style, they can `[.SomeStyle]`, and if there is a attribute to be set, they can `[SomeAttribute='abc']`, and if an option to be set they can `[%SomeProperty]`.

This solution is easier for documentation in asciidoctor-reveal.js, if not, each keyword map between reveal.js to asciidoctor-reveal.js, should be demonstrated explicitly in asciidoctor-reveal.js doc.

Since this repo is asciidoctor-reveal.js, so, probably a little bit tight couple is acceptable, : )

edl7878:
The current code may not very optimized:

section.html.slim 
```
    //reveal.js-attrib   default-value
    //data-auto-animate just set is ok
    data-auto-animate=((attr? 'auto-animate') || (option? 'auto-animate'))
    //data-auto-animate-easing	ease
    data-auto-animate-easing=((attr? 'auto-animate-easing') || (option? 'auto-animate-easing'))
    //data-auto-animate-unmatched	true
    data-auto-animate-unmatched=((attr? 'auto-animate-unmatched') || (option? 'auto-animate-unmatched'))
    //data-auto-animate-duration	1.0
    data-auto-animate-duration=((attr? 'auto-animate-duration') || (option? 'auto-animate-duration'))
    //data-auto-animate-id	absent
    data-auto-animate-id=(attr 'auto-animate-duration')
   //data-auto-animate-restart	absent
    data-auto-animate-restart=((attr? 'auto-animate-restart') || (option? 'auto-animate-restart')))
```

Mogztter:
> For now, the current modified code may not very optimized, any suggestions?

If it helps reducing code duplication, you can add an helper method in https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/helpers.rb

edl7878:
May I ask, how to export an windows standlone exe converter?

mojavelinux:
The conversation about attribute and option mapping strikes me as very off topic for this issue. I think it should be moved to a separate issue so it doesn't distract from the work being done here.

Mogztter:
> The conversation about attribute and option mapping strikes me as very off topic for this issue. I think it should be moved to a separate issue so it doesn't distract from the work being done here.

I agree we should focus on supporting reveal.js 4.x. We can create follow-up issues/pull requests to improve attributes and options mapping.

> May I ask, how to export an windows standlone exe converter?

You need to execute `npm run package`. Please note that you first need to install Node and install dependencies (using `npm i`).

I just noticed that one test is failing due to one of my change, I will fix it.

Mogztter:
@edl7878 I will also rebase, squash commits and force push.

For the next time, please create a dedicated branch using the following convention: `issue-{{issue-number}}-{{short-description}}` (in this case, it could have been: `issue-370-support-reveal-js-4-x`). Also, please use rebase (and not merge) to synchronize with the latest changes on the main branch. It makes it easier to rewrite history and squash commits.

templates/section.html.slim
```diff
@@ -63,7 +63,13 @@
     data-background-video-muted=((attr? 'background-video-muted') || (option? 'muted'))
     data-background-opacity=(attr "background-opacity")
     data-autoslide=(attr "autoslide")
-    data-state=(attr 'state'))
+    data-state=(attr 'state')
+    data-auto-animate=((attr? 'auto-animate') || (option? 'auto-animate'))
+    data-auto-animate-easing=((attr? 'auto-animate-easing') || (option? 'auto-animate-easing'))
+    data-auto-animate-unmatched=((attr? 'auto-animate-unmatched') || (option? 'auto-animate-unmatched'))
+    data-auto-animate-duration=((attr? 'auto-animate-duration') || (option? 'auto-animate-duration'))
+    data-auto-animate-id=(attr 'auto-animate-duration')
```
Mogztter, line 71:
```suggestion
    data-auto-animate-id=(attr 'auto-animate-id')
```

templates/section.html.slim
```diff
@@ -63,7 +63,13 @@
     data-background-video-muted=((attr? 'background-video-muted') || (option? 'muted'))
     data-background-opacity=(attr "background-opacity")
     data-autoslide=(attr "autoslide")
-    data-state=(attr 'state'))
+    data-state=(attr 'state')
+    data-auto-animate=((attr? 'auto-animate') || (option? 'auto-animate'))
```
Mogztter, line 67: We need to add a sample presentation in the `examples` directory to demonstrate how to use these attributes/options.

It serves multiple purposes:
- we can manually test that everything is working fine by opening the generated presentation in a browser
- we can write tests (using the existing test infrastructure) 
- we are "self-documenting" new features


I think we should also add a new page in the documentation named `auto-animate.adoc` to describe this new feature in the context of an AsciiDoc presentation.

edl7878:
> @edl7878 I will also rebase, squash commits and force push.
> 
> For the next time, please create a dedicated branch using the following convention: `issue-{{issue-number}}-{{short-description}}` (in this case, it could have been: `issue-370-support-reveal-js-4-x`). Also, please use rebase (and not merge) to synchronize with the latest changes on the main branch. It makes it easier to rewrite history and squash commits.

OK. Got it.

maxandersen:
Awesome!

PR #429, opened 2021-06-16
Mention in the doc that revealjsdir can be a CDN

maxandersen:
took me while to figure out that to actually use the standalone in a standalone manner you should use  
https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.9.0as opposed to look up using npm install etc. 

suggesting to use the cdn  location so the output is standalone and can be run and used with nothing else than the asciidoctor-revealjs binary.

### DIFF
--- a/docs/modules/setup/pages/standalone-executable.adoc
+++ b/docs/modules/setup/pages/standalone-executable.adoc
@@ -34,7 +34,7 @@ If you don't have an existing presentation, you can create a sample presentation
 [source,asciidoc]
 ----
 = Title Slide
-:revealjsdir: reveal.js
+:revealjsdir:  https://cdnjs.cloudflare.com/ajax/libs/reveal.js/3.9.0
 
 == Slide One
 
--- a/docs/modules/setup/pages/standalone-executable.adoc
+++ b/docs/modules/setup/pages/standalone-executable.adoc
@@ -34,7 +34,7 @@ If you don't have an existing presentation, you can create a sample presentation
 [source,asciidoc]
 ----
 = Title Slide
-:revealjsdir:  reveal.js
+:revealjsdir: reveal.js
 
 == Slide One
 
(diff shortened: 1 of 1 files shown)


Mogztter:
> took me while to figure out that to actually use the standalone in a standalone manner you should use
cdnjs.cloudflare.com/ajax/libs/reveal.js/3.9.0 as opposed to look up using npm install etc.

You are correct, the binary does not include reveal.js but I think you forgot this step in the "Install" section:

> Copy or clone the reveal.js presentation framework in the directory where you will build your slidedeck.Here we do a shallow clone of the repo: 
> `$ git clone -b 3.9.2 --depth 1 https://github.com/hakimel/reveal.js.git`

Using a CDN is a valid solution but it won't work if you don't have network access (wifi at conferences is not always reliable 😅).

> suggesting to use the cdn location so the output is standalone and can be run and used with nothing else than the asciidoctor-revealjs binary.

I think we can suggest it but it should be in a `TIP` admonition and it should also mention the trade-off (no Internet connection, no presentation!).

It makes me think that, in theory, we could embed reveal.js in the binary and generate a standalone HTML file (reveal.js, reveal.css... will be inlined in the HTML).
Anyway, I'm digressing a bit 😉

maxandersen:
Having a option for reveal.js dist generated would be awesome. 

Feels as such a burden to get npm init working too :)

maxandersen:
i'ved updated based on suggestions.

docs/modules/setup/pages/standalone-executable.adoc
```diff
@@ -34,7 +34,7 @@ If you don't have an existing presentation, you can create a sample presentation
 [source,asciidoc]
 ----
 = Title Slide
-:revealjsdir: reveal.js
+:revealjsdir:  reveal.js
```
Mogztter, line 37:
```suggestion
:revealjsdir: reveal.js
```

Mogztter:
Perfect thanks!

Issue #122, opened 2017-02-01
Multiple authors not taken into consideration in revealjs output?


seinecle:
asciidoc source:
(note the 2 authors)

https://raw.githubusercontent.com/seinecle/gephi-tutorials/master/src/main/asciidoc/en/plugins/twitter-streaming-importer-en.adoc

revelajs outpout:
(just one author appearing on the title page)

https://seinecle.github.io/gephi-tutorials/generated-slides/twitter-streaming-importer-en.html

obilodeau:
Confirmed that multiple authors support is not present in the slim backend: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/slim/document.html.slim#L99

Author support is completely absent in the jade backend.

obilodeau:
Jade back-end is no longer relevant in master. Removing label.

marcotama:
I would love to see this fixed. Can I help? I have a general programming background, but I am not sure where to start.

obilodeau:
Setup the Asciidoctor-revealjs plugin in development mode: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/HACKING.adoc#test-a-local-asciidoctor-revealjs-version-1

Modify the slim template here: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/document.html.slim#L106

Make sure you `bundle exec rake build` after making changes because we are using compiled slim templates now.

Relevant documentation to help you debug more: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/HACKING.adoc#inspect-the-template-system

This would be a great PR for a Hackertoberfest free t-shirt by the way, check if you haven't heard about it: https://hacktoberfest.digitalocean.com/

Ping me on https://gitter.im/asciidoctor/asciidoctor if you need interactive help. I'll try my best but can't guarantee availability.

archiloque:
I noticed there even if multiple authors are not yet supported there is an error in the original Asciidoctor file, I've opened a separated PR at https://github.com/seinecle/gephi-tutorials/pull/28

PR #437, opened 2021-08-10
resolves #31 add more complete author information

### LINKED ISSUE #31: Add more complete author information
Currently, only a single author is shown without his/her e-mail address / URL. Implement a more complete author block that shows all the authors and includes the e-mail address / URL if present.

archiloque:
Should fix https://github.com/asciidoctor/asciidoctor-reveal.js/issues/31 and https://github.com/asciidoctor/asciidoctor-reveal.js/issues/122 , I've adapted the proposed code adapted from what I saw in the html5 converter, tell me if I should adapt the style.

Among the propositions I've used a `br` to separate the authors because a `&middot;` would make the line wraps on my screen, I didn't add a custom class to the `br` because I though the `authors` class of the parent `p` should make it easy enough to customize, but I can change it if you prefer.

### DIFF
--- a/templates/title_slide.html.slim
+++ b/templates/title_slide.html.slim
@@ -24,5 +24,4 @@ section.title(class = role
   - preamble = @document.find_by context: :preamble
   - unless preamble.nil? or preamble.length == 0
     div.preamble=preamble.pop.content
-  - unless author.nil?
-    p.author: small=author
+  = generate_authors(@document)
--- a/templates/asciidoctor-compatibility.css
+++ b/templates/asciidoctor-compatibility.css
@@ -388,3 +388,13 @@ td.hdlist1 {
   font-size: 0.65em;
   margin-top: 4em;
 }
+
+.byline {
+  font-size:.8em
+}
+ul.byline {
+  list-style-type: none;
+}
+ul.byline li + li {
+  margin-top: 0.25em;
+}
--- a/templates/helpers.rb
+++ b/templates/helpers.rb
@@ -278,6 +278,42 @@ module Slim::Helpers
     end
     nil
   end
+
+  # Copied from asciidoctor/lib/asciidoctor/converter/semantic-html5.rb which is not yet shipped
+  # @todo remove this code when the new converter becomes available in the main gem
+  def generate_authors node
+    return if node.authors.empty?
+
+    if node.authors.length == 1
+      %(<p class="byline">
+#{format_author node, node.authors.first}
+</p>)
+    else
+      result = ['<ul class="byline">']
+      node.authors.each do |author|
+        result << "<li>#{format_author node, author}</li>"
+      end
+      result << '</ul>'
+      result.join Asciidoctor::LF
+    end
+  end
+
+  # Copied from asciidoctor/lib/asciidoctor/converter/semantic-html5.rb which is not yet shipped
+  # @todo remove this code when the new converter becomes available in the main gem
+  def format_author node, author
+    in_context 'author' do
+      %(<span class="author">#{node.sub_replacements author.name}#{author.email ? %( #{node.sub_macros author.email}) : ''}</span>)
+    end
+  end
+
+  # Copied from asciidoctor/lib/asciidoctor/converter/semantic-html5.rb which is not yet shipped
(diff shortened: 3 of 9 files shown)


templates/title_slide.html.slim
```diff
@@ -24,5 +24,12 @@ section.title(class = role
   - preamble = @document.find_by context: :preamble
   - unless preamble.nil? or preamble.length == 0
     div.preamble=preamble.pop.content
-  - unless author.nil?
-    p.author: small=author
+  - if @document.authors.length > 0
+    p.authors
```
Mogztter, line 28: We recently revisited the author(s) line in the new HTML 5 converter, see: https://github.com/asciidoctor/asciidoctor/blob/214b20116a7191a01238cbc4006bdec0eee8cea6/lib/asciidoctor/converter/semantic-html5.rb#L127-L142

We should probably generate the same HTML in this converter.

mojavelinux: Absolutely. Let's take this opportunity to be forward looking. It's not necesssary to cling to the older HTML.

Mogztter: We might need to add a few lines of CSS in: https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/asciidoctor-compatibility.css to make sure that it looks decent by default (i.e., when using the default/built-in style/theme).

templates/title_slide.html.slim
```diff
@@ -24,5 +24,12 @@ section.title(class = role
...
+      - @document.authors.each.with_index(1) do |author, idx|
+        - if idx > 1
+          br
+        small.author
+          span.name id="author#{idx > 1 ? idx : ''}" =sub_replacements(author.name)
+          - if author.email
+            span.email< id="email#{idx > 1 ? idx : ''}" =sub_macros(author.email)
```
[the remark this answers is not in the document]
archiloque: It was to add a whitespace before the email, but the other comment should make this code obsolete

templates/title_slide.html.slim
archiloque: If this code is already in a separated method and you try to generate semantic HTML5, I think it would be nice to try to reuse the method instead of reimplementing it, what do you think of it ? I'm not for forced reuse everywhere but in this case it seems to make sense.

Mogztter: > If this code is already in a separated method and you try to generate semantic HTML5, I think it would be nice to try to reuse the method instead of reimplementing it, what do you think of it ? 
I'm not for forced reuse everywhere but in this case it seems to make sense.

Indeed, this is a good idea and that's one of the reason why the new converter has dedicated methods.

> As the code in the main gem is not yet released maybe we can wait for it since the bugs are opened since a while

You are correct, the new converter is still at an early stage. We want to make it available sooner than later but currently it's not available in any release.
We don't have an exact date for when the new converter will be ready/available so we should not put this feature on hold. 

> or make a temporary copy with a comment indicating it should be deleted later.

Yes, I think it's totally fine to do that 👍🏻

Mogztter: > While browsing the code I've found https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/templates/helpers.rb#L54 , is the comment still relevant 


The comment is still accurate in the sense that the code was copied from https://github.com/jirutka/asciidoctor-html5s/blob/a71db48a1dd5196b668b3a3d93693c5d877c5bf3/data/templates/helpers.rb.

Both `asciidoctor-html5s` and `asciidoctor-reveal.js` are using Slim templates so it's not surprising that they are using similar helper methods.
Having said that, `asciidoctor-reveal.js` has no direct dependency on `asciidoctor-html5s`.

> (if I understand correctly asciidoctor-html5s seems to be subsumed by the new HTML 5 converter inside the main project)?

Not exactly... the goal is similar (i.e., we want to provide an HTML 5 converter that maximizes semantic constructs) but, since `asciidoctor-html5s` is a custom converter, users are free to choose which one they want to use.

archiloque: Thanks for all the explanations!

Mogztter (approved):
Looks good 👍🏻 
Thanks for your contribution!

Issue #369, opened 2020-05-27
MathJax 3 Support?


obilodeau:
With the revealjs major bump ahead. Is moving to MathJax 3 the right time?

obilodeau:
I forgot to mention that relevant information about this is in #355.

PR #443, opened 2021-08-12
Update MathJax support to version 3

archiloque:
Intended to implement https://github.com/asciidoctor/asciidoctor-reveal.js/issues/369
The configuration format changed so I adapted it using https://mathjax.github.io/MathJax-demos-web/convert-configuration/convert-configuration.html and reading the doc.

These are the documents I used for my tests: 

```
= Asciimath
:stem:

== Slide

[stem]
++++
(a + b)^2 = a^2 + 2ab + b^2
++++
```

```
= Latexmath no num
:stem: latexmath

== Slide

[stem]
++++
(a + b)^2 = a^2 + 2ab + b^2
++++

[stem]
++++
(a + b)^2 = a^2 + 2ab + b^2
++++
```

```
= Latexmath num
:stem: latexmath
:eqnums: all

== Slide

[stem]
++++
(a + b)^2 = a^2 + 2ab + b^2
++++

[stem]
++++
(a + b)^2 = a^2 + 2ab + b^2
++++
```

```
= Latexmath num ams
firstname1 middlename1 lastname1 <email1@example.net>
2017-01-31
:stem: latexmath
:eqnums: ams

== Slide

[stem]
++++
(a + b)^2 = a^2 + 2ab + b^2
++++

[stem]
++++
(a + b)^2 = a^2 + 2ab + b^2
++++
```

```
= No stem
firstname1 middlename1 lastname1 <email1@example.net>
2017-01-31

== Slide

Content
```

With this change asciidoctor-reveal.js would use MathJax 3 but asciidoctor still use MathJax 2 (see https://github.com/asciidoctor/asciidoctor/issues/3449), I don't know if having different versions is a problem.

### DIFF
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -49,22 +49,24 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     - if attr? :stem
       - eqnums_val = (attr 'eqnums', 'none')
       - eqnums_val = 'AMS' if eqnums_val == ''
-      - eqnums_opt = %( equationNumbers: { autoNumber: "#{eqnums_val}" } )
-      - mathjaxdir = (attr 'mathjaxdir', "#{cdn_base}/mathjax/2.7.6")
-      script type='text/x-mathjax-config'
-        | MathJax.Hub.Config({
-          tex2jax: {
+      - mathjaxdir = (attr 'mathjaxdir', "#{cdn_base}/mathjax/3.2.0/es5")
+      script
+        | window.MathJax = {
+          tex: {
             inlineMath: [#{Asciidoctor::INLINE_MATH_DELIMITERS[:latexmath].to_s}],
             displayMath: [#{Asciidoctor::BLOCK_MATH_DELIMITERS[:latexmath].to_s}],
-            ignoreClass: "nostem|nolatexmath"
+            tags: '#{eqnums_val}'
           },
-          asciimath2jax: {
+          options: {
+            ignoreClass: 'nostem|nolatexmath'
+          },
+          asciimath: {
             delimiters: [#{Asciidoctor::BLOCK_MATH_DELIMITERS[:asciimath].to_s}],
-            ignoreClass: "nostem|noasciimath"
           },
-          TeX: {#{eqnums_opt}}
-          });
-      script src='#{mathjaxdir}/MathJax.js?config=TeX-MML-AM_HTMLorMML'
+          loader: {
+            load: ['input/asciimath', 'output/chtml']
+          }};
+      script(async src='#{mathjaxdir}/tex-mml-chtml.js')
 
     - syntax_hl = self.syntax_highlighter
     - if syntax_hl && (syntax_hl.docinfo? :head)
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -46,28 +46,7 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
           link rel='stylesheet' href=%(#{cdn_base}/font-awesome/#{font_awesome_version}/css/v4-shims.min.css)
       - else
         link rel='stylesheet' href=(normalize_web_path %(#{attr 'iconfont-name', 'font-awesome'}.css), (attr 'stylesdir', ''), false)
-    - if attr? :stem
-      - eqnums_val = (attr 'eqnums', 'none')
-      - eqnums_val = 'AMS' if eqnums_val == ''
-      - mathjaxdir = (attr 'mathjaxdir', "#{cdn_base}/mathjax/3.2.0/es5")
-      script
-        | window.MathJax = {
-          tex: {
-            inlineMath: [#{Asciidoctor::INLINE_MATH_DELIMITERS[:latexmath].to_s}],
-            displayMath: [#{Asciidoctor::BLOCK_MATH_DELIMITERS[:latexmath].to_s}],
-            tags: '#{eqnums_val}'
-          },
-          options: {
-            ignoreClass: 'nostem|nolatexmath'
-          },
-          asciimath: {
-            delimiters: [#{Asciidoctor::BLOCK_MATH_DELIMITERS[:asciimath].to_s}],
-          },
-          loader: {
-            load: ['input/asciimath', 'output/chtml']
-          }};
-      script(async src='#{mathjaxdir}/tex-mml-chtml.js')
-
+    = generate_stem(cdn_base)
     - syntax_hl = self.syntax_highlighter
     - if syntax_hl && (syntax_hl.docinfo? :head)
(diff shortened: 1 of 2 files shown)


Mogztter:
> With this change asciidoctor-reveal.js would use MathJax 3 but asciidoctor still use MathJax 2 (see asciidoctor/asciidoctor#3449), I don't know if having different versions is a problem.

No, it should be OK. I'm already using MathJax 3 in the browser extension: https://github.com/asciidoctor/asciidoctor-browser-extension/blob/8b569beb2da80d5d020c2b006ad997c5ee79e14b/app/js/renderer.js#L52-L92 and it seems to be working just fine.

> The configuration format changed so I adapted it using mathjax.github.io/MathJax-demos-web/convert-configuration/convert-configuration.html and reading the doc.

I had to declare a `adjustDisplay` function to support STEM block.
I think we also need to define `processEscapes` since the default value changed in MathJax 3:

> The default for `processEscapes` has changed from false in version 2 to true in version 3.

http://docs.mathjax.org/en/latest/options/input/tex.html

templates/document.html.slim
```diff
@@ -49,22 +49,24 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     - if attr? :stem
       - eqnums_val = (attr 'eqnums', 'none')
       - eqnums_val = 'AMS' if eqnums_val == ''
```
Mogztter, line 51: MathJax 3 is more strict about invalid values. This is what I did in the browser extension:

https://github.com/asciidoctor/asciidoctor-browser-extension/blob/436ad4a387e4ce72a2df8bde1334bc8bc8b26efa/app/js/converter.js#L20-L28

As far as I remember, the value must be `none`, `all` or `ams` (in lowercase) otherwise an error is thrown.

archiloque: Thanks, using the code from the browser extension helped me to spot another mistake.

As the configuration values are touchy I've added some error handling, I've seen in the extension a default value is used but I think in a command-line tool raising an error is more acceptable and avoid bad surprises.

And as the logic start to become quite convoluted I moved the code to a helper. As I probably spent too much writting Java and I'm  loosy typist I love using constants so I've used some to store the possible values, but I can go back to literals if you prefer.

archiloque:
First I didn't saw there was another comment, sorry.

> I had to declare a `adjustDisplay` function to support STEM block.

I tried some tests with the files shown above and I can't see the problem appearing when I compare the result with or without the `adjustDisplay`. I wonder if it's because the HTML and CSS are different, can you check if you can reproduce the issue ? (I'm tempted to just add the code but I'd prefer to be sure if it's something that needed.

> I think we also need to define `processEscapes` since the default value changed in MathJax 3:
> 
> > The default for `processEscapes` has changed from false in version 2 to true in version 3.
> 
> http://docs.mathjax.org/en/latest/options/input/tex.html

I spotted it this morning so it's fixed.

templates/helpers.rb
```diff
@@ -314,6 +316,47 @@ def in_context name
...
+
+  # Generate the Mathjax markup to process STEM expressions
+  # @param cdn_base [String]
+  # @return [String]
+  def generate_stem(cdn_base)
+    if attr?(:stem)
+      eqnums_val = attr('eqnums', STEM_EQNUMS_NONE)
+      eqnums_val = STEM_EQNUMS_AMS if eqnums_val == ''
+      unless STEM_EQNUMS_VALID_VALUES.include?(eqnums_val)
+        raise ::ArgumentError, %(Invalid eqnums value [#{eqnums_val}], valid values are #{STEM_EQNUMS_VALID_VALUES.join(', ')})
```
mojavelinux, line 336: An Asciidoctor converter should not raise an error (unless something is truly, truly sideways). Rather, it should log a warning or error. This allows the outcome to be controlled by the user. If the user chooses to set the log failure level, then Asciidoctor will terminate with a non-zero exit code at the end.

In this case, it would be better to coerce the value to an acceptable value. We don't want to stop the conversion over such a petty problem. And we may not even need to warn.

archiloque: OK, I've changed the code to use the `ams` value and put no warning

templates/helpers.rb
```diff
@@ -314,6 +316,47 @@ def in_context name
...
+    STEM_EQNUMS_AMS,
+    'all'
+  ]
+
+  # Generate the Mathjax markup to process STEM expressions
+  # @param cdn_base [String]
+  # @return [String]
+  def generate_stem(cdn_base)
+    if attr?(:stem)
+      eqnums_val = attr('eqnums', STEM_EQNUMS_NONE)
```
Mogztter, line 333: I think we need to call `#downcase` to make sure that `:eqnums: ALL` will be correctly recognized.

templates/helpers.rb
```diff
@@ -314,6 +316,47 @@ def in_context name
...
+      eqnums_val = STEM_EQNUMS_AMS if eqnums_val == ''
```
Mogztter, line 334: Technically, I think we can remove this line, since empty `''` is not a valid value and thus will default to `STEM_EQNUMS_AMS` (in the code just below)

templates/helpers.rb
```diff
@@ -314,6 +316,47 @@ def in_context name
...
+          ignoreHtmlClass: 'nostem|nolatexmath'
+        },
+        asciimath: {
+          delimiters: [Asciidoctor::BLOCK_MATH_DELIMITERS[:asciimath]],
+        },
+        loader: {
+          load: ['input/asciimath', 'output/chtml', 'ui/menu']
+        }
+      }
+      mathjaxdir = attr('mathjaxdir', "#{cdn_base}/mathjax/3.2.0/es5")
```
Mogztter, line 355: I think we should extract MathJax version as a constant (to make it easier to bump the version).

templates/helpers.rb
archiloque: You're right, fixed!

templates/helpers.rb
archiloque: Done

templates/helpers.rb
archiloque: Done

templates/helpers.rb
```diff
@@ -314,6 +316,48 @@ def in_context name
...
+
+  STEM_EQNUMS_AMS = 'ams'
+  STEM_EQNUMS_NONE = 'none'
+  STEM_EQNUMS_VALID_VALUES = [
+    STEM_EQNUMS_NONE,
+    STEM_EQNUMS_AMS,
+    'all'
+  ]
+
+  MATHJAX_VERSION = "3.2.0"
```
Mogztter, line 328:
```suggestion
  MATHJAX_VERSION = '3.2.0'
```

Mogztter (approved):
Looks good to me 👍🏻

templates/document.html.slim
Mogztter: It's alright 👌🏻

PR #444, opened 2021-08-12
Small doc cleaning / improvement

archiloque:
No content change but small markup fixes so all code is displayed the right way, it was bothering me when I read the doc.


Mogztter:
Perfect, thanks 👌🏻

PR #445, opened 2021-08-13
Small doc markup fixes on auto-animate.adoc

archiloque:
Adding syntax highlighting, the changes on line with content is because Intellij removes end of line space characters.


Mogztter:
Thanks again 😄

PR #448, opened 2021-08-14
resolves #447 support auto-animate on code blocks

### LINKED ISSUE #447: auto-animate not working for code blocks
as discussed in #430 code blocks are not possible to auto animated like shown in https://revealjs.com/auto-animate/ - mainly because we don't have a way to assign data-id's to blocks inside a section.

Mogztter:
animate

resolves #447

### DIFF
--- a/docs/modules/converter/pages/syntax/auto-animate.adoc
+++ b/docs/modules/converter/pages/syntax/auto-animate.adoc
@@ -1,14 +1,13 @@
 = Auto Animate
 
-This feature is a wrapper of reveal.js auto animation.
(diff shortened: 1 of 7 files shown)


Mogztter:
We could avoid the visual shift on the right (from the second source block to the third and final source block) using a padding on the line number.

fixed

maxandersen:
Nice! Is this just for code Blocks or does it allow to set necessary attributes on any block ?

Mogztter:
> Nice! Is this just for code Blocks or does it allow to set necessary attributes on any block ?

This pull request add support for the auto-animate feature on source blocks (as described in the reveal.js documentation).

I didn't try on other blocks so I can't tell. What did you have in mind? Using auto animate on quote blocks?

PR #449, opened 2021-08-14
SyntaxHighlighter#docinfo? must be called after Document#content

ggrossetie:
### TODO

- [x] Add a test to make sure that Rouge stylesheet is included when `:source-highlighter: rouge` is present and when there's a least one source block.


mojavelinux:
This is the correct approach. The docinfo? method must be called after the content is converted. The same order is used by the built-in HTML converter. In that case, the converter injects the docinfo for the head position back into an assertion slot.

PR #459, opened 2021-11-22
Update commands to avoid deprecated messages

jmbruel:
To avoid the deprecated messages:
```
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set --local path '.bundle/gems'`, and stop using this flag
[DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs --all`
```

### DIFF
--- a/docs/modules/setup/pages/ruby-setup.adoc
+++ b/docs/modules/setup/pages/ruby-setup.adoc
@@ -41,8 +41,9 @@ NOTE: For some reason, when you use the system Ruby on Fedora, you also have to 
 . Install the gems into the project
 
   $ bundle config --local github.https true
-  $ bundle --path=.bundle/gems --binstubs=.bundle/.bin
-
+  $ bundle config set --local path '.bundle/gems'
+	$ bundle binstubs --all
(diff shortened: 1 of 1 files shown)


docs/modules/setup/pages/ruby-setup.adoc
```diff
@@ -41,8 +41,9 @@ NOTE: For some reason, when you use the system Ruby on Fedora, you also have to
 . Install the gems into the project
 
   $ bundle config --local github.https true
-  $ bundle --path=.bundle/gems --binstubs=.bundle/.bin
-
+  $ bundle config set --local path '.bundle/gems'
+	$ bundle binstubs --all
```
Mogztter, line 45:
```suggestion
  $ bundle binstubs --all
```

Mogztter:
I'm not a `bundler` expert but your changes seem reasonable and as far as I know `bundle binstubs` was introduced a lot time ago so we should be fine.

PR #470, opened 2022-06-01
improve Ruby setup instructions

mojavelinux:
* make prerequisites reflect current state of Ruby
* use callout to add note about gem command in Gemfile
* use bundle config to set Bundle install path
* remove unnecessary --binstubs option when calling Bundler
* use presentation.adoc as the sample presentation file name
* fix some inconsistencies with the AsciiDoc syntax

### DIFF
--- a/docs/modules/setup/pages/ruby-setup.adoc
+++ b/docs/modules/setup/pages/ruby-setup.adoc
@@ -5,47 +5,50 @@ NOTE: To ensure repeatability, we recommend that you manage your presentation pr
 
 == Prerequisites
 
-. Install {url-bundler}[Bundler^] (if not already installed) using your system's package manager or with:
+If you manage Ruby using https://rvm.io[RVM] (as recommended), make sure you switch to the default Ruby version and gemset:
+
+  $ rvm use default
+
+If you've installed Ruby using RVM, you should already have {url-bundler}[Bundler^] installed.
+You can verify this using the following command:
+
+  $ bundle -v
+
+If Bundler is not installed, you can install it using the following command:
 
   $ gem install bundler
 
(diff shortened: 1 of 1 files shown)


mojavelinux:
Once reviewed and merged, please backport this whole file to the maint-4.1.x branch.

docs/modules/setup/pages/ruby-setup.adoc
```diff
@@ -5,47 +5,50 @@ NOTE: To ensure repeatability, we recommend that you manage your presentation pr
 
 == Prerequisites
 
-. Install {url-bundler}[Bundler^] (if not already installed) using your system's package manager or with:
+If you manage Ruby using https://rvm.io[RVM] (as recommended), make sure you switch to the default Ruby version and gemset:
 
-  $ gem install bundler
+  $ rvm use default
```
Mogztter, line 10: On my machine, I always create a dedicated gemset (to avoid conflicts) using:

```
rvm use default@project-name --create
```

Should we recommend that?

mojavelinux: That's not necessary. The default gemset should never have any other gems in it other than what RVM adds. And when you switch to it `rvm use default`, you get that gemset.

In fact, I never use gemsets anymore because `bundle exec` is a much more reliable way of doing things.

mojavelinux: The default gemset should never be tainted. If you have some other gemset, then in might have other stuff in it. But that's why we are instructing the user to shift off any specialized gemset. This should return them to the default state, which is just Ruby.

> That's what bundle config --local path .bundle/gems is doing?

Absolutely. We're not using a gemset in these instructions. We are installing the gems into the project (like npm does).

PR #472, opened 2022-06-06
resolves #453 overhaul images support

### LINKED ISSUE #453: Support inline option on SVG images
When rendering asciidoc to plain html I can do e.g. `image::my-image.svg[opts=inline]` and it would inline the image into the resulting _html_.

Is is possible to do so with `asciidoctor-reveal.js`.

Cheers,
 Mario :)

ggrossetie:
resolves #453


barthel:
Not merged right now?

PR #478, opened 2022-08-21
Proofread documentation

Mogztter:


docs/modules/converter/pages/syntax/fragment.adoc
```diff
@@ -172,7 +172,7 @@ you can wrap a list block in an open block:
 
 In the above example, the title "Greek Alphabet" and all the list items will be revealed at once.
 
-And here's a another example where we apply the _step_ option on different elements:
+And here's an another example where we apply the _step_ option on different elements:
```
mojavelinux, line 175: The "a" / "an" can be dropped here entirely.

Issue #466, opened 2022-03-07
Use a consistent naming convention for Docinfo locations


Mogztter:
Currently, we support 4 locations:

<table>
<tr>
	<td>Location
    <td>Description
	<td>Docinfo suffix
<tr>
	<td><code>header</code>
    <td> At the beginning of the presentation (i.e. before all slides)
	<td><code>header-revealjs.html</code>
<tr>
	<td><code>footer</code>
    <td>At the end of the presentation (i.e. after all the slides)
	<td><code>footer-revealjs.html</code>
<tr>
	<td><code>head</code>
    <td>Just before <code>&lt;/head&gt;</code>
	<td><code>footer-revealjs.html</code>
<tr>
	<td><code>footer</code>
    <td>Just before <code>&lt;/html&gt;</code> (after <code>reveal.js</code> script)
	<td><code>footer.html</code>
</table>

In the future, we might want to inject Docinfo files at the beginning of every slide (i.e., slide header) and at the end of every slide (i.e., slide footer). Though, we might want to inject Docinfo files at the beginning of every *top-level* slide rather than on every slide... or maybe only on *vertical* (i.e. secondary/vertical navigation) slides.

#### Proposal

- `header`: Just before and outside of `<div class="slides">` (to accomodate https://github.com/asciidoctor/asciidoctor-reveal.js/issues/376)
- `first`:  At the beginning of the presentation (i.e. before all slides)
- `last`: At the end of the presentation (i.e. after all the slides)
- `head`: Just before `</head>`
- `footer`: Just before `</html>`

//cc @mojavelinux

Mogztter:
@mojavelinux I would greatly appreciate your input on this proposal since this is the last standing issue before the 5.0.0 release and I'm a bit unsure about the terminology...

I'm also questioning if we should have Docinfo with and without suffix? For instance, `head` and `footer` location are not tied to reveal.js since we insert content respectively just before `</head>` and just before `</html>`. But at the same time, it might be useful to dissociate docinfo files that are specific to a reveal.js presentation...

yann-soubeyrand:
Hello,

[...]

PR #510, opened 2023-06-24
Add header docinfo support

yann-soubeyrand:
Head, header and footer reveal.js docinfo, and footer docinfo are already supported. However, header docinfo is not, but may be useful to add a logo for example. It can also confuse users that this is not working like in other AsciiDoc documents (maybe head docinfo should be added as well).

### DIFF
--- a/templates/document.html.slim
+++ b/templates/document.html.slim
@@ -66,6 +66,8 @@ html lang=(attr :lang, 'en' unless attr? :nolang)
     - unless (_docinfo = docinfo :head, '-revealjs.html').empty?
       =_docinfo
   body
+    - unless (docinfo_content = (docinfo :header, '.html')).empty?
+      =docinfo_content
     .reveal
       / Any section element inside of this container is displayed as a slide
       .slides


ggrossetie:
I would love your opinion on https://github.com/asciidoctor/asciidoctor-reveal.js/issues/466
Once we agree on which use cases are supported we can move forward and add the missing docinfo locations.

>  but may be useful to add a logo for example

The logo will appear on each slides? before the first slide? Could you please provide a simple example?

yann-soubeyrand:
Hi @ggrossetie, sorry for the late reply. You can find an example here: https://github.com/camptocamp/asciidoctor-revealjs-template/blob/bf5d7c1cf9a91c6a9eb76a5200ff700270d34a56/docinfo-footer.html#L1-L6. There’s a `presentation.html` in this repo showing the result. It took me some time to figure out why I couldn’t get my logos when putting them in the `docinfo-header.html` file and I was surprised to discover that `docinfo-footer.html` was supported, but not the other `docinfo-*.html`. I can’t put my logos in `docinfo-header-revealjs.html`, because their position is wrong in this case.

yann-soubeyrand:
As an (unrelated) side note, Font Awesome v6 seems to work with v4 shims: https://github.com/camptocamp/asciidoctor-revealjs-template/blob/example/css/font-awesome.css.

PR #518, opened 2023-07-18
Add support for fragments on table rows

johthor:
This change will add stepping through table rows via option, role or attribute `rowSteps` on the table block.


ggrossetie:
Superseeded by #586 

The codebase has shifted quite a bit, and the feature needed implementation across both JS and Ruby sides to be complete. I ended up refactoring your idea to fit the current state of things, which meant your PR needed to be reworked entirely.
Thank you for starting this work 🙌🏻

PR #529, opened 2024-04-13
Using [%step] on colist works the same as on ulist

ibethus:
Hi,

This PR aims to modify the behavior of the `step` attribute on a callout block. I believe callouts should behave consistently with traditional lists when prefixed with this attribute. 
This adjustment can enhance the presentation of code fragments, offering a more interactive experience.

**Example :**
```asciidoc
[source, javascript]
----
let hello = "hello world"; <1>
console.log(hello); <2>
----
[%step]
<1> First line
<2> Second line
```

In the current implementation, only one fragment would be created. With this modification, individual fragments will be generated for each callout line, aligning with the intended behavior.

**Code :**
On the code side, I've only repositioned the fragment class definition in the colist template and included an example in the `step.adoc` file to illustrate this modification.

Cheers!

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -153,6 +153,7 @@ namespace :examples do
     FileUtils.cp 'examples/a11y-dark.css', "#{PUBLIC_DIR}/a11y-dark.css"
     FileUtils.cp 'examples/release-5.1.html', "#{PUBLIC_DIR}/release-5.1.html"
     FileUtils.cp 'examples/release-5.1.css', "#{PUBLIC_DIR}/release-5.1.css"
+    FileUtils.cp 'examples/release-5.2.html', "#{PUBLIC_DIR}/release-5.2.html"
   end
 end
 
--- a/examples/release-5.2.adoc
+++ b/examples/release-5.2.adoc
@@ -1,0 +1,35 @@
+= Asciidoctor reveal.js 5.2
+//:stem:
+:source-highlighter: highlight.js
+:highlightjs-theme: a11y-dark.css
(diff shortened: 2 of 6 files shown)


ggrossetie:
Thanks!
It would be great if you could create a `release-5.2.adoc` in https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/examples (similar to https://github.com/asciidoctor/asciidoctor-reveal.js/blob/master/examples/release-5.1.adoc).

Then, copy the HTML file in the public directory:

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/cbe2ac7f72877050ce15ee4a38fd54ca91590a78/Rakefile#L154

And add a link in:

https://github.com/asciidoctor/asciidoctor-reveal.js/blob/cbe2ac7f72877050ce15ee4a38fd54ca91590a78/src/index.html#L24

That way, we can see the result using Netlify preview: https://deploy-preview-529--asciidoctor-revealjs-examples.netlify.app/

ggrossetie:
One test is failing because you've enabled `:icons: font` and update the template. You will need to update the expected result `test/doctest/steps.html`

ibethus:
Thanks for the feedback!

Documentation and test case updated, hope this matches your expectations :+1:

ggrossetie:
Yes that's awesome, thanks 🎉 
If anyone is curious, you can see the result at: https://deploy-preview-529--asciidoctor-revealjs-examples.netlify.app/release-5.2#/_iterating_through_callouts

mojavelinux:
That's really nice!

PR #531, opened 2024-04-23
Update syntax-highlighting.adoc to say there can be no blank lines after the title

kito99:


docs/modules/converter/pages/syntax/syntax-highlighting.adoc
```diff
@@ -11,6 +11,10 @@ You can activate Highlight.js syntax highlighting (disabled by default) by setti
 :source-highlighter: highlight.js
 ----
 
+[NOTE]
+====
+There can be no blank lines between the title and the `source-highlighter` attribute. If you insert a blank line, the source highlighter will be ignored.
+
```
ggrossetie, line 17:
```suggestion
[IMPORTANT]
====
The first empty line in the document marks the end of the header.
In other words, if you declare the `source-highlighter` attribute after an empty line it won't activate syntax highlighting.

https://docs.asciidoctor.org/asciidoc/latest/document/header/#when-does-the-document-header-end
====
```

